PR #5, opened 2014-12-18
element library -> nuclide library

elliottbiondo:
This script adds entries to the ALARA elelib.std file for each nuclide within the file as if it were a pure substance. This allows users to create matlib entries using nuclides of their choosing without having to create custom entries in the elelib.


gonuke:
What about a short test?  And we might as well put a full nuclib in the repo.

elliottbiondo:
ping @gonuke

gonuke:
nuclib should probably live in the `data` directory instead; see `elelib.std`

elliottbiondo:
Fixed @gonuke.

elliottbiondo:
ping @gonuke

gonuke:
Thanks @elliottbiondo

PR #6, opened 2015-08-11
integrate_energy

elliottbiondo:
This PR adds the optional "integrate_energy" modifier to the output block. This modifier changes the way groupwise photon emission density is calculated, in order to preserve total power (e.g. MeV/s). A more detailed description of this is found in ```doc/inputtext.html``` and the header of ```sample/sample9```. An additional sample file was created in order to test this functionality. Results were validated with hand calculations (expected values are given in the sample file). This PR should be backwards-compatible with all existing ALARA input files.


gonuke:
Looks good @elliottbiondo - Thanks!

PR #9, opened 2016-05-03
fix compiler warnings

rwcarlsen:


gonuke:
Thanks!  I must not be keeping you busy enough :)

rwcarlsen:
Pretty easy fixes (about 30 min of work) - just did what the compiler told
me.  Should fix some pretty serious memory leak issues among other things.
But fixes like this probably just make it even less likely ALARA ever gets
its long overdue rewrite :-)

elliottbiondo:
These fixes don't seem to break anything on a full-scale problem.

gonuke:
And the closest thing we have to tests are the sample problems, for which we diff the outputs.

It passes there - I'll merge.

PR #12, opened 2016-09-09
Fix build errors when using newer versions of GCC

ljacobson64:
I was getting various build errors when using GCC 5 and 6. I tested that these changes work on GCC 6.2.


gonuke:
Thanks @ljacobson64

PR #13, opened 2016-10-31
Small fix to elemen name length.

Pshriwise:
Limits element name lengths to 2. Keeps ALARA from looking for string characters in places where there aren't any. This crops up due to a problem with the way strlen is interpreted for non-null-terminated strings in C.


gonuke:
Is this the thing we fixed in Feb while doing the ITER SDR calc?  I thought we merged that already, but don't see it in code...

Pshriwise:
@gonuke yup. I'm not sure. I think we used this branch for most of that calc. @makeclean asked me to PR it this afternoon.

PR #17, opened 2018-02-08
add compile into gitignore

zxkjack123:
Major change:
1. Add `compile` in the `.gitignore`. The `compile` is generated from the `autoreconf`. Ignore it makes repo cleaner.


gonuke:
Thanks @zxkjack123

PR #18, opened 2018-02-09
Update confiugre doc

zxkjack123:
Major change:
1. Add `autoreconf -fi` command before `configure`.


gonuke:
Thanks for this improvement @zxkjack123

PR #23, opened 2019-07-29
Alara documentation in sphinx

bradlaufenberg:
This PR includes the rest of the pages of the ALARA documentation transferred over to rst format in order to build using Sphinx. One slight problem are the links to various documents/data files at the end of `installguide.rst` in the data-dir section. I'm unsure where/how I should copy these files since many are not plain text files.


gonuke:
Thanks @bradlaufenberg - I'm going to merge this and then we can look at specific issues to update/improve what's here.

PR #27, opened 2019-07-31
Redid glossary to use sphinx auto-glossary suppoort

bradlaufenberg:
Rewrote `glossary.rst` and redid all user guide file's links to the glossary to now use the Sphinx engine auto glossary. Tested the build successfully using the `doc.Makefile`.

### DIFF
--- a/doc/usersguide/Manual_syntax.rst
+++ b/doc/usersguide/Manual_syntax.rst
@@ -3,9 +3,6 @@ Manual Syntax
 =============
 
 +---------------------------------------+-----------------------------------------------------------+
-| [G]		                        |This icon indicates that a definition exists in the        |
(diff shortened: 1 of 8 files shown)


gonuke:
Since this glossary automatically creates links, I think we can get rid of all the `[G]` entries

bradlaufenberg:
Should I get rid of the glossary section in the Manual syntax page as well? All it indicates is the '[G]' means the word is in the glossary, and the next section says that text written in red is a link to another section in the manual.

bradlaufenberg:
I ended up removing it (see the latest commit ), will be an easy fix if you want me to put it back.

gonuke:
Thanks @bradlaufenberg

PR #28, opened 2019-08-05
Further improving ALARA documentation

bradlaufenberg:
These are the changes we talked about in our meeting earlier today. Everything was successfully built with the `doc.Makefile` when I tested it.


gonuke:
Thanks @bradlaufenberg

PR #29, opened 2019-08-19
Improved the User guide's toc

bradlaufenberg:


bradlaufenberg:
Was able to successfully build website after changes were made.

gonuke:
Looking at what you did to fix this, I think see the issue.  For consistency, it might make sense to update all the headings where top-level headings have a single *, second level headings have 2, and so on...

bradlaufenberg:
For Sphinx one * inclosing words means it will italicize it, while ** inclosing words means that it will bold those words. The reason I chose to bold multiple headings opposed to underlining it and making it subtitles was to make sure the TOC wasn't too big and cluttered. So, would you still like me to go back and do what you proposed in your last comment?

gonuke:
Gotcha... I was mixing up my markdown syntax.

I do think I want things marked as headings.  You can nest different levels of headings with different characters used to underline (maybe you knew that already):

```
Level 1
======

Level 2
-----------

Level 3
~~~~~~~
```

And then you can provide directives to the `toctree` regarding how many levels to include

bradlaufenberg:
I made the changes to the headings like you requested. The only other thing we may want to edit is which headings appear in the toc for each of the pages of the users guide, (this will probably be whatever you prefer if you just want to relay it to me I can go in and make the changes).

gonuke:
Good questions - I'm not sure how best to concisely articulate my preference....

Maybe we should sit down and discuss it somewhat.

bradlaufenberg:
Sounds like a plan. I'm free every day next week so whenever you are available would work for me.

bradlaufenberg:
The merge I did for myself in the above commit was how I updated my master branch with the edits from this PR after my original `user_gude_devo` branch became messed up and I had to delete it.

gonuke:
This looks a lot better!  I just noticed that the keywords in the input file syntax are now mixed case: some are capitalized and some are lower case.  It's probably better to have them all lower case.

bradlaufenberg:
Just made that small change to the input file keywords. Let me know if there is anything else you would like me to do for this PR.

gonuke:
Thanks @bradlaufenberg - I think that takes care of it!

Issue #30, opened 2019-10-09
ALARA compilation problem


kyfranzen:
Hi,
When trying to install ALARA following the README instructions I get the following error after the make command:
ake[3]: Leaving directory `/localhome/restricted/ken/alara/bld/src/Util'
make[3]: Entering directory `/localhome/restricted/ken/alara/bld/src'
make[3]: Nothing to be done for `all-am'.
make[3]: Leaving directory `/localhome/restricted/ken/alara/bld/src'
make[2]: Leaving directory `/localhome/restricted/ken/alara/bld/src'
Making all in tools
make[2]: Entering directory `/localhome/restricted/ken/alara/bld/tools'
make[2]: Nothing to be done for `all'.
make[2]: Leaving directory `/localhome/restricted/ken/alara/bld/tools'
Making all in doc
make[2]: Entering directory `/localhome/restricted/ken/alara/bld/doc'
make[2]: *** No rule to make target `commandtext.html', needed by `all-am'.  Stop.
make[2]: Leaving directory `/localhome/restricted/ken/alara/bld/doc'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/localhome/restricted/ken/alara/bld'
make: *** [all] Error 2
(py27) [ken@attila bld]$

Any ideas on what to do?

Thanks!,
Ken Franzen

ljacobson64:
I get the same error. It appears to have been caused by #22.  I'm not entirely sure how to fix it at this time

gonuke:
Thanks for reporting this!  We've been updating the way that we do documentation, but didn't consider how it was integrated with the build system.  I've pushed a fix in #31

ljacobson64:
@kyfranzen can you try again? @gonuke fixed this issue in #31.

kyfranzen:
I successfully tried to install the latest master branch.
Nicely done!
Thanks,
Ken

PR #31, opened 2019-10-13
Fix doc build

gonuke:
Given recent changes in how documentation is provided, this fixes the build system to install appropriate RST files instead of the prior HTML files.

### DIFF
--- a/Makefile.h
+++ b/Makefile.h
@@ -3,6 +3,7 @@
 
 # where to put documentation, data and sample input/output
 docdir = $(prefix)/doc
+ugdir = $(docdir)/usersguide
 nonxsdir = $(prefix)/data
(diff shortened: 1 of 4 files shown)


ljacobson64:
I tested this PR and the build issue seen in #30 no longer occurs.

I see that a `doc` directory appears in the install prefix directory, and it contains various `.rst` files. However in the `doc` directory inside the repo, there are both `.rst` files and `.html` files. What exactly is going on here?

gonuke:
We're replacing the HTML documentation with RST versions.  I kept the HTML around to make sure we weren't missing anything.

kyfranzen:
I also tested the new fix and it seems to be compiling OK now.
Thanks for looking into this!
Ken

PR #32, opened 2020-01-15
Flatten file hierarchy

bradlaufenberg:
This is the first step in switching the build process for ALARA from automake to CMake. Here the file hierarchy is  simplified by having less sub-directories. There were then changes to `src/Makefile.am` in order to take the change in location of the files into account. Each individual file was then scanned through to change the pathways referenced since all files are now found in the `src` directory.


gonuke:
This doesn't build for me - have you tested it?

bradlaufenberg:
The latest commit allows ALARA to be successfully built.

gonuke:
LGTM - I can build it successfully and the tests seem to pass

PR #35, opened 2020-06-03
Switch to CMake

bradlaufenberg:
So far I have successfully built the `src` directory using CMake. The next step is to do the same for the rest of the sub directories.


src/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# src directory CMake file
+
+include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+configure_file(version.h.in "${CMAKE_CURRENT_SOURCE_DIR}/config.h")
+
+FILE(GLOB h_files "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
+FILE(GLOB c_files "${CMAKE_CURRENT_SOURCE_DIR}/*.C")
+SET(EXECUTABLES ${c_files})
```
gonuke, line 8:
```suggestion
SET(SOURCE_FILES ${c_files})
```

src/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+
+# Add the Data Library
+add_subdirectory(DataLib)
+
+# Add all source files
+add_executable(alara ${EXECUTABLES})
```
gonuke, line 14: to match the change above
```suggestion
add_executable(alara ${SOURCE_FILES})
```

src/input.C
```diff
@@ -1,7 +1,7 @@
 /* $Id: input.C,v 1.14 2007-12-20 22:03:56 wilsonp Exp $ */
 #include "alara.h"
 #include "input_tokens.h"
-#include "dflt_datadir.h"
+//#include "dflt_datadir.h"
```
gonuke, line 4: Have we successfully figured out how to generate this file with CMake as we were doing with autotools?

gonuke: I think we have put this info in `config.h` instead (that is current generated from `version.h.in`).  We can probably delete this line.

bradlaufenberg:
@gonuke  I believe everything is done but before I push just wanted to clarify two things from the latest comments you had made.

1) Is the symbolic link between `src/alara.h` and `src/DataLib/alara.h` the target_link_libraries line in `src/CMakeLists.txt`  ?

2) Do I set the `xsdir` and `nonxsdir` in `src/CmakeLists.txt` ?

gonuke:
The symbolic link looks like an extra file that appears in the repository.  It's actually not directly related to your changes, but I think that your changes make it less necessary.

You should be able to set the value of `xsdir` and `nonxsdir` in `src/CMakeLists.txt` (or equally valid may be `CMakeLists.txt`) based on the known installation location and relative paths from there.

src/CMakeLists.txt
```diff
@@ -12,4 +11,7 @@ add_subdirectory(DataLib)
 
 # Add all source files
 add_executable(alara ${SOURCE_FILES})
-target_link_libraries(alara DataLib)
+
+# Set direcotry variable names
```
gonuke, line 15: typo
```suggestion
# Set directory variable names
```

src/CMakeLists.txt
```diff
@@ -12,4 +11,7 @@ add_subdirectory(DataLib)
...
+set(DFLT_XSDIR ${xsdir})
```
gonuke, line 16: I don't think this will do what it needs to do - we should discuss (see note in Slack)

src/DataLib/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# CMake file in the DataLib direcotry within the src directory
+
+FILE(GLOB h_files "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
```
gonuke, line 3: Not sure that you need `h_files` defined.

PR #37, opened 2021-03-28
Add CMake build system plus other goodies

ljacobson64:
This main content of this PR is adding the CMake build system. Other additions with this PR include:

- Adding a script that runs all of the sample problems
- Adding a script that diffs the sample problem output against assumed-to-be-correct output
- Updating the readme to reflect the CMake build system and the new sample file running procedure
- Updating the install guide (there is some duplicated information between this file and the readme; perhaps information should only be in one place or the other?)
- Adding documentation showing how to build the documentation
- Changing the source branch when building the documentation from master to main
- Fixing a common typo where "opened" is misspelled as "openned"


gonuke (changes requested):
Thanks for pushing this across the finish line.  Just a question about what should be tested.

ljacobson64:
@gonuke Did you mean to make more posts here? Not sure what you're asking

sample/run_samples.sh
```diff
@@ -1,50 +1,21 @@
...
-echo sample9
-sed -e "s/\.\/sample9.photonSrc_int/$outdir\/sample9.photonSrc_int/" sample9 > sample9.tmp
-sed -i "s/\.\/sample9.photonSrc/$outdir\/sample9.photonSrc/" sample9.tmp
-$alarabin -v 3 -t $outdir/sample9.tree sample9.tmp > $outdir/sample9.out
-rm sample9.tmp
+# Run samples
+samples=sample?
+for sample in ${samples}; do
+  echo ${sample}
+  ${ALARA} -v 2 -t output/${sample}.tree ${sample} > output/${sample}.out
```
gonuke, line 20: Testing at verbosity 3 is useful to make sure the tree is being built correctly, but makes diffs harder.  

We could diff the trees instead?

gonuke (commented):
Not sure what happened to this comment the first time

sample/run_samples.sh
ljacobson64: Perhaps I can come up with a way to have it skip checking the lines that contain the timing info when doing the diffs

ljacobson64: I changed it back to verbosity 3, but I also changed the code so that the timing results don't show up until verbosity level 4. Is that acceptable? Another option would be to create a new command line flag that removes all timing results.

gonuke (approved):

gonuke:
Thanks @ljacobson64

PR #56, opened 2024-07-18
Reading MT data

### LINKED ISSUE #43: Read list of reaction type (MT) data
Read data about different reaction types (ENDF MT numbers) from a CSV file.

eitan-weinstein:
Closes #43

Creates a function that is exclusively used to read in the `mt_table.csv` and store the data in a nested dictionary with the MT numbers as the keys and the only item being another dictionary containing the reaction type. This structure will allow for the dictionary to be accessed again in other functions that will tally the types and counts of emitted particles and calculate the change in KZA for each reaction type.


gonuke:
This looks good @eitan-weinstein - let's add another script that will grow into the top level "driver" that will import an run this method, perhaps `process_fendl3.2.py`

eitan-weinstein:
Sounds good, @gonuke. Thanks for taking a look at this so quickly.

gonuke (approved):
Thanks @eitan-weinstein

Issue #44, opened 2024-07-17
Create a count of emitted particles for each reaction type


gonuke:
Using the string description of emitted particles, process it to determine how many of each particle type are emitted.

gonuke:
Store this information in a dictionary of form `{particle : number}`

MicahGale:
I'm wondering if the GNDS data format has this information in an easy to interpret way. This might also be helpful for #43.

gonuke:
Include a test that shows that this works for a few different cases

PR #57, opened 2024-07-18
Calculating nucleon changes

### LINKED ISSUE #45: Calculate the change in nucleons for a given reaction
Given the number of emitted particles of each type, determine what the change in nucleons is for a given reaction and store this information.  Include isomeric changes.  Store the result as an integer in the form of the KZA encoding.

eitan-weinstein:
Closes #45.

Creates a method for calculating the change in neutrons and protons in response to neutron activation and subsequent decay for a given reaction type. This function can be used to process the reaction types from the dictionary produced from reading in `mt_table.csv`.


gonuke:
Probably makes sense to tackle #44 first since you need that info to use #45

gonuke (approved):
This looks good to me - thanks @eitan-weinstein

Issue #55, opened 2024-07-17
Pre-process all reactions for relevant data


gonuke:
Use results generated in #44 and #45 to preprocess all reactions for relevant data.

gonuke:
Include a test that shows this works for a handful of real MT data that is included in the file

PR #58, opened 2024-07-18
Counting numbers of emitted particles for a given reaction.

### LINKED ISSUE #44: Create a count of emitted particles for each reaction type

eitan-weinstein:
Closes #44.

Establishes a method for identifying and tallying the different emitted particles for a specific reaction as defined by the MT data from `mt_table.csv`.


src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -14,6 +14,67 @@
...
+            number_str = emitted_particle_string[i] + number_str
+        else:
+            break
+
+    if number_str:
+        particle_count = int(number_str)
+    elif particle in emitted_particle_string:
+        particle_count = 1
+    else:
+        particle_count = None
```
gonuke, line 49: It might make sense to have this return 0 for this case.  Then you can still use the math without any concern.

eitan-weinstein:
I think that it might make sense to do all of the testing for #44, #45, and #55 in one because #55 will combine the functionality of the nucleon changes and the emitted particle counting anyways, so the relevant testing would go hand in hand. Let me know what you think @gonuke because I could also certainly go back and make separate testing PRs for #44 and #45 before moving on.

src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -14,6 +14,67 @@
...
+            the product. If particles not present, returns None rather than 0.
+    """
+
+    particle_index = emitted_particle_string.find(particle)
+    number_str = ''
+    for i in range(particle_index - 1, -1, -1):
+        if emitted_particle_string[i].isdigit():
+            number_str = emitted_particle_string[i] + number_str
+        else:
+            break
```
gonuke, line 42: Whenever I see a `for` loop with a `break` like this, I wonder if it's better to use a `while` loop:
```suggestion
    number_str = ''
    particle_count_index = emitted_particle_string.find(particle) - 1
    while emitted_particle_string[particle_count_index].isdigit():
        number_str = emitted_particle_string[particle_count_index] + number_str
        particle_count_index -= 1
```

gonuke (changes requested):
A couple of changes requested here.

gonuke:
> I think that it might make sense to do all of the testing for #44, #45, and #55 in one because #55 will combine the functionality of the nucleon changes and the emitted particle counting anyways, so the relevant testing would go hand in hand. Let me know what you think @gonuke because I could also certainly go back and make separate testing PRs for #44 and #45 before moving on.

Good point - I forgot my own request for testing.  It would have been good to have testing as part of #57 that just tested that method by itself.  I think it would be good to add `pytest` tests here that test the individual methods separate from the data in the MT CSV file.  Let me know if you want/need tips on that.

eitan-weinstein:
> > I think that it might make sense to do all of the testing for #44, #45, and #55 in one because #55 will combine the functionality of the nucleon changes and the emitted particle counting anyways, so the relevant testing would go hand in hand. Let me know what you think @gonuke because I could also certainly go back and make separate testing PRs for #44 and #45 before moving on.
> 
> Good point - I forgot my own request for testing. It would have been good to have testing as part of #57 that just tested that method by itself. I think it would be good to add `pytest` tests here that test the individual methods separate from the data in the MT CSV file. Let me know if you want/need tips on that.

Alright, that make sense. I have not used `pytest` before, so I would love to hear whatever tips you might have to offer before I jump in.

gonuke:
> If/when you think it makes sense, you can just merge it and it becomes part of your branch and PR for you to build upon

Note: if you merge on github, you'll need to fetch/pull this branch to your local machine to have those changes represented there

eitan-weinstein:
> > If/when you think it makes sense, you can just merge it and it becomes part of your branch and PR for you to build upon
> 
> Note: if you merge on github, you'll need to fetch/pull this branch to your local machine to have those changes represented there

Yep, I got it. Thanks @gonuke

src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -14,6 +14,68 @@
...
+        particle_count (int): Count of the target particle present in
+            the product.
+    """
+
+    num_str = ''
+    particle_count_index = emitted_particle_string.find(particle) - 1
+    if particle_count_index < 0:
+        particle_count = 0
+    else:
+        while emitted_particle_string[particle_count_index].isdigit():
```
gonuke, line 41: Good catch, but I think this should work?
```suggestion
    while particle_count_index >=0 and emitted_particle_string[particle_count_index].isdigit():
```

gonuke (changes requested):
Streamlining of while loop

src/DataLib/fendl32B_retrofit/reaction_data.py
eitan-weinstein: Yeah, that's definitely nicer than how I had set it up. I'll double check that it works and throw it in.

src/DataLib/fendl32B_retrofit/tests/test_reaction_data.py
```diff
@@ -0,0 +1,19 @@
+
+# Import packages
+import pytest
+import reaction_data as rxd
```
gonuke, line 4: Since this is in a different directory, this import might fail now.  I recommend putting it in the same directory for now, and we can worry about directory structure another time.

src/DataLib/fendl32B_retrofit/test_reaction_data.py
gonuke, line 1: This file is now redundant with the file in the `tests` directory.

gonuke:
Perhaps you're already working on it, but now you can add some tests for `nucleon_change()` and possibly (but slightly less importantly) `emission_breakdown()`

gonuke:
Are you planning more here?  I was thinking:
* [ ] tests for `nucleon_change()`
* [ ] tests for `emission_breakdown()`

eitan-weinstein:
> Are you planning more here? I was thinking:
> 
> * [ ]  tests for `nucleon_change()`
> * [ ]  tests for `emission_breakdown()`
> 
> as part of this PR since those methods are in place. Good to get the tests in before we move on.

Sorry, I misunderstood and had been planning to put those in another PR, but have mostly finished out writing those tests, so I can copy them over to this branch and push another commit with them shortly so that all of those tests will be in this PR.

src/DataLib/fendl32B_retrofit/test_reaction_data.py
```diff
@@ -0,0 +1,53 @@
...
+    obs = rxd.count_emitted_particles(particle, emitted_particle_string)
+    assert obs == exp
+
+@pytest.mark.parametrize(
+    "(emitted_particles, exp)",
+    [
+        ('np' , {'n' : 1, 'p' : 1}),
+        ('2d3t3He', {'d' : 2, 't' : 3, '3He' : 1}),
+        ('total', {}),
+        ('α75γn3', {'α' : 1, 'γ': 75, 'n': 1})
```
gonuke, line 30: missing comma
```suggestion
        ('α75γn3', {'α' : 1, 'γ': 75, 'n': 1}),
```

src/DataLib/fendl32B_retrofit/test_reaction_data.py
```diff
@@ -0,0 +1,53 @@
+
+# Import packages
+import pytest
+import reaction_data as rxd
+from numpy import array
+
+@pytest.mark.parametrize(
+        "(particle, emitted_particle_string, exp)",
```
gonuke, line 8: syntax:
```suggestion
        "particle, emitted_particle_string, exp",
```

src/DataLib/fendl32B_retrofit/test_reaction_data.py
```diff
@@ -0,0 +1,53 @@
...
+        ('3He'    , '3He8α'                , 1  ) ,
+        ('α'      , '3He8α'                , 8  )
+    ]
+)
+def test_count_emitted_particles(particle, emitted_particle_string, exp):
+    obs = rxd.count_emitted_particles(particle, emitted_particle_string)
+    assert obs == exp
+
+@pytest.mark.parametrize(
+    "(emitted_particles, exp)",
```
gonuke, line 25: syntax:
```suggestion
    "emitted_particles, exp",
```

src/DataLib/fendl32B_retrofit/test_reaction_data.py
```diff
@@ -0,0 +1,53 @@
...
+        ('Xα', {})
+    ]
+)
+def test_emission_breakdown(emitted_particles, exp):
+    obs = rxd.emission_breakdown(emitted_particles)
+    assert obs == exp
+
+
+@pytest.mark.parametrize(
+    "(emission_dict, exp)",
```
gonuke, line 40: syntax:
```suggestion
    "emission_dict, exp",
```

gonuke (changes requested):
I found some syntax errors when I tried to run these tests.  You should be able to test by calling `pytest` at the command-line.

Also, you'll need to find the right way to compare arrays.

gonuke:
Take a look at: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html

eitan-weinstein:
> I found some syntax errors when I tried to run these tests. You should be able to test by calling `pytest` at the command-line.
> 
> Also, you'll need to find the right way to compare arrays.

Thanks for the guidance @gonuke. `pytest` is passing now.

src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -14,6 +14,74 @@
...
+        emission_dict (dict): Dictionary containing each individual particle
+            type and their respetive counts. For an np emission, this would
+            read out as {'n': 1, 'p': 1}.
+    """
+
+    emission_dict = {
+        particle: count_emitted_particles(particle, emitted_particles)
+        for particle in NP_dict.keys()
+        if particle in emitted_particles and all(
+            spec_case not in emitted_particles for spec_case in spec_reactions
```
gonuke, line 79: It seems we should definitely discuss this condition and whether the logic we are using is causing us to more work than necessary....

gonuke (commented):
This is looking pretty good, but I think we need to discuss the logic of dealing with `spec_reactions`

src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -14,6 +14,74 @@
...
+    Returns:
+        emission_dict (dict): Dictionary containing each individual particle
+            type and their respetive counts. For an np emission, this would
+            read out as {'n': 1, 'p': 1}.
+    """
+
+    emission_dict = {
+        particle: count_emitted_particles(particle, emitted_particles)
+        for particle in NP_dict.keys()
+        if particle in emitted_particles and all(
```
gonuke, line 78: Perhaps this instead...  it's not critical, but does better convey what is important and technically will be more efficient (although in this use case that's largely irrelevant)

```suggestion
        if particle in emitted_particles and not any(
```

gonuke (changes requested):
I think this is the last change before we merge

gonuke (approved):

PR #59, opened 2024-07-19
MT data preprocessing, including change in KZAs

### LINKED ISSUE #55: Pre-process all reactions for relevant data

eitan-weinstein:
Closes #55.

Modifies `reaction_data.py` to extract all of the data from an already-read dictionary of the loaded `mt_table.csv`. This uses the methods from #44 and #45 to find and tally the particles emitted from a given reaction and calculate the respective change in KZA value.

This includes testing for the function `process_mt_data()`, which relies on six csv files containing single MT data directly from ENDF manual ([https://www.oecd-nea.org/dbdata/data/manual-endf/endf102_MT.pdf](url)). These files, in a subdirectory titled `files_for_tests` are included in this PR.


gonuke:
Can you rebase on the recently merged #44  to show a more limited change set?

gonuke:
I don't think that did what you expected...

eitan-weinstein:
> I don't think that did what you expected...

No, it did not. I'm having some trouble with this rebase, sorry for all of the extraneous messages and commits that are coming through while I try to sort this out.

src/DataLib/fendl32B_retrofit/test_reaction_data.py
```diff
@@ -50,4 +50,44 @@ def test_emission_breakdown(emitted_particles, exp):
 )
 def test_nucleon_changes(emission_dict, exp):
     obs = rxd.nucleon_changes(emission_dict)
-    assert_array_equal(obs, exp)
\ No newline at end of file
+    assert_array_equal(obs, exp)
+
+@pytest.mark.parametrize(
+    "mt_data_csv, exp",
```
gonuke, line 56: This is good use of parametrization, but I'm not sure they need to be in different files.  Let's discuss what needs to be tested specifically here that is not tested elsewhere. It needs to test that you can read/process multiple entries. But the processing of each individual entry is already demonstrated in other tests, so it may not be necessary to have that breadth here.  Perhaps a case with a single entry and a case with multiple entries? 

Let me know if you think I'm missing something.

gonuke (changes requested):
Thanks @eitan-weinstein - A few small changes, but mostly in good shape.

src/DataLib/fendl32B_retrofit/test_reaction_data.py
eitan-weinstein: > This is good use of parametrization, but I'm not sure they need to be in different files. Let's discuss what needs to be tested specifically here that is not tested elsewhere. It needs to test that you can read/process multiple entries. But the processing of each individual entry is already demonstrated in other tests, so it may not be necessary to have that breadth here. Perhaps a case with a single entry and a case with multiple entries?
> 
> Let me know if you think I'm missing something.

No, that makes sense. I was not sure on the convention for how many tests to include, so that's why I had multiple files, but in principle, they are not accomplishing different things.

src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -135,4 +138,72 @@ def load_mt_table(csv_path):
...
+    """
+
+    for MT, data in mt_dict.items():
+        emitted_particles = data['Reaction'].split(',')[1][:-1]
+        emission_dict = emission_breakdown(emitted_particles)
+        change_N, change_P = nucleon_changes(emission_dict)
+        M = check_for_isomer(emitted_particles)
+
+        # Conditionally remove isomer tags from emitted particle strings
+        if emitted_particles.rfind(str(M)) > 0:
```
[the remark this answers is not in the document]
eitan-weinstein: > Can't you just check the integer value of `M`?

I realize I could also just do `if M > 0:` because either way, M is going to be an integer, but regardless this still simplifies things.

src/DataLib/fendl32B_retrofit/test_reaction_data.py
gonuke: There is definitely an art testing.  Normally, I don't worry too. much about "too many" tests, but I would like to limit the number of testing files we need

eitan-weinstein:
It looks like the changes for the empty `mt_table` case are taking me a little longer than I thought, so I'll just get that in to you on Monday morning.

src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -98,9 +98,12 @@ def nucleon_changes(emission_dict):
             change in neutrons and protons in a nucleus as a result of neutron
             activation and subsequent decay. The array is in the format of
             array([neutron_change, proton_change]).
+
+            If emission_dict is empty, NP_change will be returned as an array
+            of None values, in the format array([None, None]).
     """
     
-    NP_change = array([None , None])
+    NP_change = array([None, None])
```
gonuke, line 106: I think this can just be `None` as default:
```suggestion
    NP_change = None
```

src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -134,5 +137,76 @@ def load_mt_table(csv_path):
...
+                    {'Reaction'         :               (z, emission)},
+                    {'delKZA'           :       integer change in KZA},
+                    {'Emitted Particles : string of emitted particles}
+            }
+    """
+
+    for MT, data in mt_dict.items():
+        emitted_particles = data['Reaction'].split(',')[1][:-1]
+        emission_dict = emission_breakdown(emitted_particles)
+        change_NP = list(nucleon_changes(emission_dict))
```
gonuke, line 198: I don't think you need to save this as a list, with the above change, it's either a numpy array or `None`
```suggestion
        change_NP = nucleon_changes(emission_dict)
```

src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -134,5 +137,76 @@ def load_mt_table(csv_path):
...
+        M = check_for_isomer(emitted_particles)
+
+        # Conditionally remove isomer tags from emitted particle strings
+        if emitted_particles.rfind(str(M)) > 0:
+            emitted_particles = emitted_particles[:-len(str(M))]
+        
+        if change_NP != [None, None]:
+            change_N, change_P = change_NP
+            data['delKZA'] = (change_P * 1000 + change_P + change_N) * 10 + M
```
gonuke, line 207:
```suggestion
        if change_NP:
            data['delKZA'] = (change_NP[0] * 1000 + change_NP[0] + change_NP[1]) * 10 + M
```

gonuke (changes requested):
A little clarification on my intent for streamlining the nucleon change info

src/DataLib/fendl32B_retrofit/reaction_data.py
eitan-weinstein: Ok, I understand what you mean now, but I think the syntax needed to make this work would be `if change_NP is not None:` because otherwise it can yield this message:

`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

gonuke:
Thanks for pulling this together @eitan-weinstein

PR #60, opened 2024-07-25
Extracting MT data from ENDF files.

### LINKED ISSUE #46: Determine list of reaction types in a given file
For a single ENDF file, representing a single nuclide, determine the list of reaction types that exist in file MF=3.

eitan-weinstein:
Closes #46.

Adds a new Python file `tendl_preprocessing.py`, which includes a function to extract the material ID and MT numbers from a given TENDL (ENDF) file.

Also includes testing for this new function, with test files contained in the `files_for_tests` subdirectory.


gonuke:
I'll review this following a rebase/merge

eitan-weinstein:
I just rebased it now. I think I will also quickly modify the test files, as per our discussion with the last PR to avoid redundant files.

src/DataLib/fendl32B_retrofit/tendl_preprocessing.py
```diff
@@ -0,0 +1,27 @@
...
+    matb = tape.material_numbers[0]
+    # Set MF for cross sections
+    xs_MF = 3
+    file = tape.material(matb).file(xs_MF)
+    # Extract the MT numbers that are present in the file
+    MTs = [MT.MT for MT in file.sections.to_list()]
+
+    endf_specs = [matb, MTs]
+    
+    return endf_specs
```
gonuke, line 27: I'm not sure if it's important to formally create a list of this data or if you could just do the following:
```suggestion
        return matb, MTs
```

gonuke: It's not super important, but these decisions sometimes imply specific design needs/choices, and wondering if that's true here.

gonuke (changes requested):
This looks pretty good - one minor design question - probably not critical.

src/DataLib/fendl32B_retrofit/tendl_preprocessing.py
eitan-weinstein: There's not any specific reason why I create the list beforehand, I can remove it to simplify things.

gonuke:
This looks good, but I now need to install `ENDFtk` to test it myself.  I'm going to trust that the tests work for now to help move this forward, but will try to get my environment setup correctly soon.

gonuke (approved):

PR #61, opened 2024-07-25
Establish static template for NJOY/GROUPR input file

### LINKED ISSUE #49: Write a template string for the basic GROUPR input
Formulate a static string that can be used as a template for running GROUPR on different nuclides and reaction types.

eitan-weinstein:
Closes #49.

Creates a new file `groupr_tools.py` where the GROUPR/NJOY methods will be written, starting with a method that creates a static template for the NJOY input file containing all of the constant values. This template is constructed as a dictionary, containing "cards," in the parlance of NJOY documentation. Isotope and file specific data will be added through a separate method to this dictionary.


gonuke:
I think we should look at string Templates for this: https://docs.python.org/3.4/library/string.html#template-strings

The NJOY/GROUPR input is essentially static except for a couple of simple things, so it might be easiest to just define the whole file in a template and substitute in those things.

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -0,0 +1,87 @@
...
+    Returns:
+        input_template (string.Template): A static string template containing
+            the general structure of the NJOY input file to run GROUPR,
+            $identifiers for substitutable items
+            (https://docs.python.org/3.4/library/string.html#template-strings)
+        """
+
+    input_string = 'groupr\n'
+    for card_num, card in cards.items():
+        input_string += format_card(card_num, card)
```
gonuke, line 83: I think we can perhaps hard-code the full template, either with the numerical constants included or via template:
```
"""groupr
$NENDF $NPEND $NGOUT1 $NGOUT2
$mat_id $IGN $IGG $IWT $LORD $NTEMP $NSIGZ $IPRINT
(etc)
"""
```
 and then replace the standard variables in on step, and the other variables as needed.

This would avoid the arcane logic you need tin `format_card()`

gonuke (changes requested):
More guidance on the ideas of using `Template`

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -0,0 +1,36 @@
...
+NSIGZ = 1 # number of sigma zeroes (default=1)
+IPRINT = 1 # long print option (0/1=minimum/maximum) --(default=1)
+ISMOOTH = 1 # switch on/off smoother operation (1/0, default=1=on)
+TEMP = 293.16 # temperature in Kelvin
+SIGZ = 0 # sigma zero values (including infinity)
+MFD = 3 # file to be processed
+MATD = 0 # next mat number to be processed
+
+# Create input file general template
+njoy_input = Template(
```
gonuke, line 23: I don't know the answer, but wonder if you could do the substitution of the fixed items above immediately, and separately do substitution of the variable stuff later?

That is, can substitutions be done in two rounds?

gonuke (changes requested):
Great simplification here... one little question to make life easier down the road

src/DataLib/fendl32B_retrofit/groupr_tools.py
eitan-weinstein: I believe it can, but I'll test it out now and if it does work, push a new commit with that modification.

eitan-weinstein: Ok, I just made a new update that substitutes in the constants immediately. I had to use `safe_substitute` here because not all of the placeholders are being called, which produces a `str` instead of substituting back into a `Template`, so I have to convert back to a `Template` again, but it seems to work fine.

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -0,0 +1,43 @@
...
+     0/
+    stop
+    """
+).safe_substitute(
+    NENDF = NENDF, NPEND = NPEND, NGOUT1 = NGOUT1, NGOUT2 = NGOUT2, # Card 1
+    IGN = IGN, IGG = IGG, IWT = IWT, LORD = LORD,                   # Card 2
+    NTEMP = NTEMP, NSIGZ = NSIGZ, IPRINT = IPRINT,                  # Card 2
+    TEMP = TEMP,                                                    # Card 4
+    SIGZ = SIGZ,                                                    # Card 5
+    MATD = MATD                                                     # Card 10
```
gonuke, line 42: Now that I see this, it makes me think you can just define the numerical values right here, with the comments you use above here instead.

That is, instead of defining the values of the variables above and using them here as "X = X", you can just define the values here in the middle of the substitution.

gonuke (changes requested):

src/DataLib/fendl32B_retrofit/groupr_tools.py
eitan-weinstein: I could, the only reason that I had the values defined separately was for the comment lines explaining what each of them was for. If you think that is not necessary, I can pull it out, or if you think that the explanations would be better fit somewhere else, I'd be happy to modify it.

gonuke: I like the comments.  I just wonder whether you can basically cut/paste that into the substitution arguments, with the comments...

eitan-weinstein: Ok, I'll modify it and you can let me know if it fits what you're looking for.

gonuke (approved):
Excellent!

PR #62, opened 2024-07-25
Filling in the NJOY input string.Template

eitan-weinstein:
Closes #54.

Creates a method to substitute in material-specific values in the place of fillers in the static string.Template that stores the NJOY input file string.

PR #63, opened 2024-07-26
Running NJOY for GROUPR

### LINKED ISSUE #50: Write a method that invokes NJOY to perform GROUPR conversion
Create a method that invokes NJOY through a subprocess to process endf/pendf files into gendf multi-group data.

eitan-weinstein:
Closes #50.

Creates a method that uses the `NJOY` input file formatted and created in #62 and executes an `NJOY` run using `subprocess`. Moreover, this includes a method called `ensure_gendf_markers`, which writes in additional lines to the processed GENDF files that the `NJOY` run does not otherwise produce, but are necessary to avoid warning messages that pop up when reading in the GENDF files with `ENDFtk`.


src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -107,4 +108,95 @@ def write_njoy_input_file(template):
...
+
+    # Define the input files
+    INPUT = 'groupr.inp'
+    OUTPUT = 'groupr.out'
+
+    # Run NJOY
+    result = subprocess.run(['njoy'], input=open(INPUT).read(),
+                            text=  True, capture_output=True)
+    with open(OUTPUT, 'w') as output_file:
+        output_file.write(result.stdout)
```
gonuke, line 190: can't you just set `stdout=output_file` in the `subprocess` call?

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -107,4 +108,95 @@ def write_njoy_input_file(template):
...
+
+    # If the run is successful, log out the output
+    # and make a copy of the file as a .GENDF file
+    if not result.stderr:
+        output = subprocess.run(['cat', 'output'],
+                                capture_output=True, text = True)
```
gonuke, line 196: I don't see where you use this `output` variable? Even if you did, isn't it easier to just read the file?

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -107,4 +108,95 @@ def write_njoy_input_file(template):
...
+
+        gendf_path = f'tendl_2017_{element}{str(A).zfill(3)}.gendf'
+        subprocess.run(['cp', 'tape31', gendf_path])
```
gonuke, line 199: There are python commands to copy and/or move a file.  You probably want to move this file not copy it.  e.g. `Path.rename()`

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -107,4 +108,95 @@ def write_njoy_input_file(template):
...
+
+    updates = [
+        (mf3_identifier, mf3_FEND_RECORD),
+        (mf1_identifier, mf1_SEND_RECORD)
+    ]
+
+    for identifier, new_line in updates:
+        last_line_index = file_str.rfind(identifier)
+        line_number = file_str[:last_line_index].count('\n')
+        file_lines.insert(line_number + 1, new_line)
```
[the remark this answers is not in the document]
gonuke: We might also consider a string `replace` or regular expression for this

gonuke (changes requested):
A few suggestions for more pythonic results.

eitan-weinstein:
Thank you for the suggestions @gonuke. Let me know how this looks to you now.

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -110,16 +112,57 @@ def write_njoy_input_file(template):
...
 
+def insert_after_identifier(file_str, identifier, new_line):
+    """
+    Insert a new line immediately after the last occurrence of a line
+    containing the specified identifier in the file string.
+
+    Arguments:
+        file_str (str): The content of the GENDF file as a string.
+        identifier (str): The identifier to search for in the file content.
+        new_line (str): The new line to insert after the identified line.
```
gonuke, line 123: might be a better variable name than `new_line`.  Perhaps `record_for_insertion` or something since `new_line` means something specific

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -110,16 +112,57 @@ def write_njoy_input_file(template):
...
     """
     Edit the GENDF files produced from an NJOY GROUPR run to include file and
-    section records that are not automatically written out to the file by
-    NJOY. Missing these records will not cause errors, but will raise
-    messages when they are read by ENDFtk, which expects these markers, so
-    this method ensures that they are present. Edits will only be made if
-    the SEND record for MF1 or the FEND record for MF3 are not present.
-    The formatting for these records can be found at:
-    https://t2.lanl.gov/nis/endf/intro06.html
+        section records that are not automatically written out to the file by
```
gonuke, line 144: Looks like a change in indentation to take care of here.

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -110,16 +112,57 @@ def write_njoy_input_file(template):
     with open('groupr.inp', 'w') as f:
```
gonuke, line 112: noticing this filename is hardcoded in different places, so might want to define this in a single place and pass it around or assume it's global.

Also - this method may be too simple and clear to need a standalone method rather than just include the code where you need it.

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -184,19 +222,15 @@ def run_njoy(element, A, matb):
     OUTPUT = 'groupr.out'
```
gonuke, line 222: Do you use this output file anywhere?

gonuke (changes requested):
A good step forward - a few improvements recommended here.

src/DataLib/fendl32B_retrofit/groupr_tools.py
eitan-weinstein: No, I can take it out.

src/DataLib/fendl32B_retrofit/groupr_tools.py
eitan-weinstein: That's actually a question I had; I think I have been formatting my docstrings like this in general. If this indentation should not be there, I can go back and change all of them, but if you would prefer that be in a different PR for the functions that have already been merged, I can go do that too.

gonuke (approved):
This all looks good

PR #64, opened 2024-07-31
Cleaning up files produced for or during the NJOY run.

### LINKED ISSUE #53: Clean up temporary files
Write methods to clean up any temporary files generated during the process.

eitan-weinstein:
Closes #53.

Creates a method to delete the `NJOY` input file and the "tapes" that it processes (i.e. the starting ENDF and PENDF files and the converted GENDF file). The GENDF file saved under the name `tape31` can be safely deleted because another copy of the file is saved separately in the function `run_njoy` with a more descriptive name anyways.

This method also renames the automated `NJOY` output file.


src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -107,4 +108,113 @@ def write_njoy_input_file(template):
...
+            same directory as all other saved files from the script run.
+    
+    Returns:
+        None
+    """
+
+    njoy_files = [f'groupr.inp', 'tape20', 'tape21', 'tape31']
+    for file in njoy_files:
+        subprocess.run(['rm', file])
+    subprocess.run(['mv', 'output', output_path])
```
gonuke, line 220: This is a bad way to remove files.  See `Path.unlink()`

gonuke (changes requested):
Calling a subprocess is a blunt and overbearing way to delete files

PR #65, opened 2024-07-31
Extracting the parent KZA values from GENDF files.

### LINKED ISSUE #47: Extract the parent nuclide description from a GENDF file
Extract the nuclide description from a GENDF file and reformat into the KZA integer format for ALARA.

eitan-weinstein:
Closes #47.

Creates a method to read a GENDF file and extract the the KZA value of the parent isotope and return it as an integer.


gonuke (approved):
This looks good - thanks @eitan-weinstein

PR #67, opened 2024-08-02
Extracting cross section data from GENDF files

### LINKED ISSUE #48: Extract cross-sections from GENDF file
Extract and parse the cross-section information from a GENDF file for a given reaction type

eitan-weinstein:
Closes #48.

Creates a method for extracting the cross section data from a GENDF file by MT number and another method to iterate through all of the given MT numbers in the GENDF file to call this method.

Makes small modifications to `extract_endf_specs()` to make it more versatile and usable for a GENDF file, which needs to be done so that the updated MT values are used (some MT numbers present in the preprocessed ENDF/PENDF files do not remain after processing with GROUPR).

### DIFF
--- a/src/DataLib/fendl32B_retrofit/groupr_tools.py
+++ b/src/DataLib/fendl32B_retrofit/groupr_tools.py
@@ -91,7 +91,7 @@ def fill_input_template(material_id, MTs, element, A, mt_dict):
 
     card9_lines = []
     MFD = 3 # ENDF file tag for cross-section data
-    for MT in MTs & mt_dict.keys():
+    for MT in MTs:
         mtname = mt_dict[MT]['Reaction']
         card9_lines.append(f'{MFD} {MT} "{mtname}" /') 
     card9 = '\n '.join(card9_lines)
--- a/src/DataLib/fendl32B_retrofit/groupr_tools.py
+++ b/src/DataLib/fendl32B_retrofit/groupr_tools.py
@@ -91,9 +91,10 @@ def fill_input_template(material_id, MTs, element, A, mt_dict):
 
     card9_lines = []
     for MT in MTs:
-        mtname = mt_dict[MT]['Reaction']
-        MFD = 3 # ENDF file tag for cross-section data
-        card9_lines.append(f'{MFD} {MT} "{mtname}" /') 
+        if MT in mt_dict.keys():
+            mtname = mt_dict[MT]['Reaction']
+            MFD = 3 # ENDF file tag for cross-section data
+            card9_lines.append(f'{MFD} {MT} "{mtname}" /') 
     card9 = '\n '.join(card9_lines)
     return njoy_input.substitute(
         mat_id=material_id,
--- a/src/DataLib/fendl32B_retrofit/groupr_tools.py
+++ b/src/DataLib/fendl32B_retrofit/groupr_tools.py
@@ -90,11 +90,10 @@ def fill_input_template(material_id, MTs, element, A, mt_dict):
     title = f'"{Z}-{element}-{A} for TENDL 2017"'
 
     card9_lines = []
-    for MT in MTs:
-        if MT in mt_dict.keys():
-            mtname = mt_dict[MT]['Reaction']
-            MFD = 3 # ENDF file tag for cross-section data
-            card9_lines.append(f'{MFD} {MT} "{mtname}" /') 
+    MFD = 3 # ENDF file tag for cross-section data
+    for MT in MTs & mt_dict.keys():
+        mtname = mt_dict[MT]['Reaction']
+        card9_lines.append(f'{MFD} {MT} "{mtname}" /') 
     card9 = '\n '.join(card9_lines)
     return njoy_input.substitute(
         mat_id=material_id,
--- a/src/DataLib/fendl32B_retrofit/process_fendl3.2.py
+++ b/src/DataLib/fendl32B_retrofit/process_fendl3.2.py
@@ -20,6 +20,10 @@ def main():
     groupr_tools.cleanup_njoy_files()
 
     pKZA = tp.extract_gendf_pkza(gendf_path)
+    # Extract MT values again from GENDF file as there may be some difference
+    # from the original MT values in the ENDF/PENDF files
+    endftk_file_obj, MTs = tp.extract_endf_specs(gendf_path, 'GENDF')
+    gendf_data = tp.iterate_MTs(MTs, endftk_file_obj, mt_dict, pKZA)
 
 if __name__ == '__main__':
     main()
--- a/src/DataLib/fendl32B_retrofit/process_fendl3.2.py
+++ b/src/DataLib/fendl32B_retrofit/process_fendl3.2.py
@@ -14,6 +14,7 @@ def main():
     pendf_path = 'tape21'
 
     material_id, MTs, endftk_file_obj = tp.extract_endf_specs(endf_path)
+    MTs = list(set(MTs).intersection(mt_dict.keys()))
     njoy_input = groupr_tools.fill_input_template(material_id, MTs, 'Fe', 56, mt_dict)
     groupr_tools.write_njoy_input_file(njoy_input)
     gendf_path = groupr_tools.run_njoy('Fe', 56, material_id)
--- a/src/DataLib/fendl32B_retrofit/process_fendl3.2.py
+++ b/src/DataLib/fendl32B_retrofit/process_fendl3.2.py
@@ -13,17 +13,19 @@ def main():
     endf_path = 'tape20'
     pendf_path = 'tape21'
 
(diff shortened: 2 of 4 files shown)


src/DataLib/fendl32B_retrofit/tendl_processing.py
```diff
@@ -49,4 +58,81 @@ def extract_gendf_pkza(gendf_path):
...
+            dKZA = 'N/A'
+        else:
+            dKZA = pKZA - mt_dict[MT]['delKZA']
+        emitted_particles = mt_dict[MT]['Emitted Particles']
+        cross_sections_by_MT.append(sigma_list)
+        dKZAs.append(dKZA)
+        emitted_particles_list.append(emitted_particles)
+        groups.append(len(sigma_list))
+
+    gendf_data = pd.DataFrame({
```
gonuke, line 130: curious to see how this data frame looks.  I thought you needed to have a list of dictionaries to initialize it this way, or use `from_dict()`:
image

gonuke (changes requested):
We probably need to discuss what we do with MTs that have no daughter KZA.

src/DataLib/fendl32B_retrofit/tendl_processing.py
```diff
@@ -49,4 +58,81 @@ def extract_gendf_pkza(gendf_path):
...
+    """
+
+    cross_sections_by_MT    =    []
+    emitted_particles_list  =    []
+    dKZAs                   =    []
+    groups                  =    []
+
+    for MT in MTs:
+        sigma_list = extract_cross_sections(file_obj, MT)
+        if mt_dict[MT]['delKZA'] == 'N/A':
```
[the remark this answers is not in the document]
eitan-weinstein: I'm happy to exclude these, but I'm just wondering if there is any utility in keeping the "Total" reaction. There may not be, or we could sum all of the other cross sections to recalculate that list, but I just wanted to check on that beforehand.

src/DataLib/fendl32B_retrofit/tendl_processing.py
gonuke: Although I am now questioning whether the calculated daughters match the emitted particle lists - back to the tests...

src/DataLib/fendl32B_retrofit/tendl_processing.py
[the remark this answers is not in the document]
gonuke: Do you know if the "total" reaction only includes reaction channels that result in a different final nuclide?  For example, does it include elastic scattering?  If so, then it's not very useful to us for activation.

src/DataLib/fendl32B_retrofit/tendl_processing.py
gonuke: Or maybe this is a sign that it's not working??? I don't know why an emitted particle of 2n in row 4 results in a nuclide that is one nucleon heavier?  And rows 11-20 indicate an emission of n but have a nuclide one lighter (and various isomers)

eitan-weinstein: You're right and I hadn't noticed this before. I'm guessing if it is calculating incorrectly, our issue is probably coming from how we are tracking the particles, so I'll go back and look through that function to see if there is something that we missed there.

src/DataLib/fendl32B_retrofit/tendl_processing.py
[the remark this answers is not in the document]
eitan-weinstein: From what I've read in the ENDF-6 manual, it looks like the total does include scattering as well, so we can remove it.

src/DataLib/fendl32B_retrofit/tendl_processing.py
gonuke: The tests for that seem to be capturing that correctly, though...  You may want to debug by looking at the data before and after you put it into the pandas dataframe and/or read the CSV

eitan-weinstein: Ok, I'll check that out.

eitan-weinstein: So far it seems that the daughter KZAs are not changed at all from how they are calculated in the for-loop to how they appear in the Data Frame, but I'll keep digging.

eitan-weinstein: I think I got it! In line `118` where I calculate the dKZA, it is currently written as `dKZA = pKZA - mt_dict[MT]['delKZA']`, but the way that we included the delKZA values in the dictionary, they are all negative, so really it should be `dKZA = pKZA + mt_dict[MT]['delKZA']`, which seems to answer the question of why we are seeing impossible nucleon math.

src/DataLib/fendl32B_retrofit/reaction_data.py
```diff
@@ -192,7 +192,7 @@ def process_mt_data(mt_dict):
             }
     """
 
-    for MT, data in mt_dict.items():
+    for MT, data in list(mt_dict.items()):
```
gonuke, line 195: Why does this need to be a `list`?

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -91,9 +91,10 @@ def fill_input_template(material_id, MTs, element, A, mt_dict):
 
     card9_lines = []
     for MT in MTs:
```
gonuke, line 93: I think you can use set algebra here to find the intersection of `MTs` and `mt_dict.keys()` and avoid the if statement.
```suggestion
    for MT in {MTs} & mt_dict.keys():
```

src/DataLib/fendl32B_retrofit/groupr_tools.py
```diff
@@ -91,9 +91,10 @@ def fill_input_template(material_id, MTs, element, A, mt_dict):
...
-        mtname = mt_dict[MT]['Reaction']
-        MFD = 3 # ENDF file tag for cross-section data
-        card9_lines.append(f'{MFD} {MT} "{mtname}" /') 
+        if MT in mt_dict.keys():
+            mtname = mt_dict[MT]['Reaction']
+            MFD = 3 # ENDF file tag for cross-section data
```
gonuke, line 96: This can be set outside the loop since it doesn't change as you go through the loop

src/DataLib/fendl32B_retrofit/process_fendl3.2.py
```diff
@@ -13,13 +13,19 @@ def main():
     endf_path = 'tape20'
     pendf_path = 'tape21'
 
-    material_id, MTs = tp.extract_endf_specs(endf_path)
+    material_id, MTs, endftk_file_obj = tp.extract_endf_specs(endf_path)
```
gonuke, line 16: could you find the intersection of `MTs` and `mt_dict.keys()` here to make your life simpler?

gonuke (changes requested):
Thanks for sleuthing out that error.  I wonder if we can make a test for that???  it may not be necessary...

src/DataLib/fendl32B_retrofit/reaction_data.py
eitan-weinstein: I read that iterating directly through the `items` in a `dict` and using those to delete `keys` from the `dict` can cause runtime errors, so by creating a separate `list`, any potential errors caused by self-references within the `dict` can be avoided.

src/DataLib/fendl32B_retrofit/process_fendl3.2.py
eitan-weinstein: I suppose I could, although I'm not sure that it offers much benefit as the intersection is found in line 94 of `groupr_tools.py`, with the line `for MT in MTs & mt_dict.keys():`, which will only pass the acceptable MT values for the NJOY run, so only MT values in that intersection will appear in the processed GENDF file.

gonuke: The benefit is that you restrict the set of MT numbers as soon as possible.  This can prevent confusion about which MT numbers are included in the list at which points in the code.  If you had a reason to keep the full list of MT numbers for some other operation, that would support making this down-selection as late as possible.

src/DataLib/fendl32B_retrofit/process_fendl3.2.py
```diff
@@ -13,13 +13,20 @@ def main():
...
+    MTs = list(set(MTs).intersection(mt_dict.keys()))
```
gonuke, line 17: This doesn't need to be a `list` after this set operation.  Forcing it to be a list is not wrong, but it conveys that a list is required.  (I won't hold up the merge for this)

gonuke (approved):
This looks good enough now, thank @eitan-weinstein

PR #69, opened 2024-08-07
Fixing order of cross section data for FENDL3.2b retrofit

eitan-weinstein:
Reverses the order of the `sigma_list` list in `tendl_processing.py` for the FENDL3.2b retrofit. After comparing the data produced through the GROUPR processing in this retrofit with FENDL2.0 data, the shapes of the plots were flipped, implying that the way that the cross sections were being saved in the retrofit was in reverse order. Upon reversing the order of the list, the two plots matched very closely. 

I will include a full comparison script either in this PR or another, depending on your preference @gonuke, but either way, this seems to be a necessary fix regardless.


gonuke (approved):
Thanks for the review

PR #70, opened 2024-08-08
Comparison of FENDL3.2b with FENDL2.0 for select isotopes and reactions.

eitan-weinstein:
Includes a python script that processes cross section data from FENDL3.2b and FENDL2.0 for a specific isotope and reaction to plot the cross sections over each other against the 175 energy groups in the Vitamin-J structure. Additionally, a number of example cases are included.


gonuke (changes requested):
I can't find a specific code change to suggest (not enough time/attention) but the threshold reactions are at the wrong energy, probably because of the order in which you prepend with 0 and then assign the energies.  I think the simple solution may be just to append with 0 instead of prepending.

src/DataLib/fendl32B_retrofit/fendl2_comparison/fendl_comparison.py
```diff
@@ -0,0 +1,289 @@
...
+   Returns:
+      cross_sections (list): Potentially modified list of cross sections, with
+         zeroes prepended to bring the length of the list to 175.
+   """
+   
+   current_length = len(cross_sections)
+   zeros_needed = 175 - current_length
+   cross_sections = [0] * zeros_needed + cross_sections
+
+   return cross_sections
```
gonuke, line 94: I think this should append:
```suggestion
def append_with_zeros(cross_sections):
   """
   Include zeroes into the end of a list such that the total length of
      of the list is 175, corresponding to the Vitamin-J group structure.
   
   Arguments:
      cross_sections (list): List of cross sections, of length less than or
         equal to 175.
   
   Returns:
      cross_sections (list): Potentially modified list of cross sections, with
         zeroes appended to bring the length of the list to 175.
   """
   
   current_length = len(cross_sections)
   zeros_needed = 175 - current_length
   cross_sections = cross_sections + [0] * zeros_needed

   return cross_sections
```

gonuke (changes requested):
I proposed specific changes for the prepend that I think will make all the threshold reactions look better

gonuke:
I wonder if @bohmt has any thoughts on these cross-section comparison plots?

MicahGale:
@gonuke, and @eitan-weinstein, you may have discussed this while I was out, but I'm not a fan of adding nuclear data and plots (binary data) to git due to bloating the repo, and the principle of git is for "inputs" not "outputs"

As for the plots:

1. Could you export as `svg` so that way we can zoom in with vector graphics?
2. Some of these (specifically Co-59 (n,g)) look "out of phase". Are you using energy bin edges consistently for plotting (i.e., lower, upper, middle)? Or did they check the nuclear data? I think checking the resonance parameters would be a quick check of that. I don't remember where those are stored though. I feel like it would `MF=1` though.

PR #71, opened 2025-01-15
Modify ALARA Input File syntax for added emphasis 

anu1217:
Suggesting some minor wording-related changes that may help draw reader attention to code features that would cause an ALARA run to fail.


gonuke (approved):
Thanks @anu1217 - these all look good to me

PR #72, opened 2025-03-05
Set C++11 as C++ Standard

giovanni-mariano:
On some systems, cmake assumes by default the C++17 standard for compilation. However, the dynamic exception specifications are not allowed in the C++17 standard (and deprecated since C++11).

The scope of the commit is to set the C++11 standard in the CMakeLists.txt file.


MicahGale:
@giovanni-mariano thanks for doing this. This will help make the code more robust. Long term though I think modernizing is the solution. 

@gonuke should we open an issue to remove these specifications, and in general move to a newer C++ version?

PR #75, opened 2025-08-27
Including ENDF-6 MT Descriptions

eitan-weinstein:
MT reference table is currently referenced in a number of places and is adapted in mt_table.csv from a URL. I think it might be a good idea to include a PDF version here in case the URL is changed or the file is taken down. Not strictly necessary for the code operation, but it might be valuable for documentation. @gonuke let me know if you think this is a worthwhile addition.


gonuke:
If we're going to include this, it may be useful to 
a. keep at least the title page for reference (or maybe just the whole report)
b. put it a different directory like maybe the developer-info folder

PR #76, opened 2025-09-05
Integrating FENDL3.2b Preprocessing with ALARA data processing.

eitan-weinstein:
This PR is aimed at integrating the Python preprocessing program contained in `ALARA/src/DataLib/fendl32B_retrofit/` with the other data libraries in `DataLib/`. The essential logic of this is accomplished with the creation of a C++ file pair, `FENDL3.cpp` and `FENDL3.h` that collect the necessary parent KZA, daughter KZA, and cross sections broken down by reaction pathways, modeled after other ALARA data libraries, with special attention focused on basing this new code off of `IEAFLib.C`.

In order to procure this data, `DATALIB_FENDL3` is treated like any other library option that can be input into an ALARA run, by including its functionality in `DataLib.C` and `DataLib.h` (as well as documentation in the other data library format files to track the update and keep documentation consistent across files). However, because the data requires pre-processing through `ALARA/src/DataLib/fendl32B_retrofit/process_fendl3.2.py`, `FENDL3.cpp` includes functionality to execute this Python script automatically and access the reformatted data directly. To do so, a number of minor modifications were needed for these Python scripts to allow for them to be called from a different directory. This is because they were otherwise structured such that all files that would be referenced, created, modified, etc. would all be in the same directory from which the script was being executed (`ALARA/src/DataLib/fendl32B_retrofit/`). As such, in maintaining the separation of the directory as nested within DataLib (which I still think makes sense organizationally to avoid clutter in DataLib), modifications were made to make `process_fendl3.2.py` more versatile to be called either from `DataLib/` or `DataLib/fendl32B_retrofit`.

### DIFF
--- a/src/DataLib/ALARAJOY_wrapper/groupr_tools.py
+++ b/src/DataLib/ALARAJOY_wrapper/groupr_tools.py
@@ -3,7 +3,7 @@ from string import Template
 import subprocess
 from pathlib import Path
 import re
-from os import getcwd
+import os
 
 def set_directory():
     '''
@@ -18,7 +18,7 @@ def set_directory():
             command was called.
     '''
 
-    dir = getcwd()
+    dir = os.getcwd()
     fendl_dir = 'ALARAJOY_wrapper'
     if fendl_dir not in dir:
         dir += f'/{fendl_dir}' # Should only be called from one directory up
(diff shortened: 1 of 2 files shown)


gonuke:
Off the top, I'm going to ask you to choose a name other than FENDL3 for this. What makes it different from the others is not the source of the data, but the way it's being processed.  (To be fair, some of the previous names may also violate this idea, but might as well start changing it now.) 

Obvious names aren't coming to mind immediately, but maybe something like ALARAJOY to suggest that it's our own wrapper to NJOY....??? Happy for other alternatives.

PR #94, opened 2025-09-16
Updating ALARAJOY Python Wrapper

eitan-weinstein:
Following #93, this PR makes minor adjustments to the ALARAJOYWrapper used to process FENDL3 nuclear data to a usable format for ALARA. An overview of the changes is catalogued as such:

1. Expanding versatility for `preprocess_fendl3.py` to both be run from other directories, as well as more control for organized file saving and access. For small numbers of isotopes being processed, this might not be strictly necessary, but for processing the whole TENDL2017 neutron activation database, better file control seemed like a worthwhile modification.
2. Renaming certain particles: When I originally wrote this program, I did not reference the specific naming conventions for alpha, gamma, and 3He particles in ALARA. Updates were made to match these to ALARA's expectations (α --> a, γ --> g, 3He --> h).

### DIFF
--- a/tools/ALARAJOYWrapper/tendl_processing.py
+++ b/tools/ALARAJOYWrapper/tendl_processing.py
@@ -105,8 +105,9 @@ def extract_gendf_pkza(gendf_path):
     Z, element, A = first_line.split('-')[:3]
 
     Z = int(Z)
+    isomer_tag = 'm' if 'm' in A.lower() else 'n'
     M = 2 if 'n' in A.lower() else 1 if 'm' in A.lower() else 0
-    A = int(A.lower().split(' ')[0].split('m')[0])
+    A = int(A.lower().split(' ')[0].split(isomer_tag)[0])
     pKZA = (Z * 1000 + A) * 10 + M
     return pKZA
 


eitan-weinstein:
I've also written a Python script that uses `BeautifulSoup` to search through the TENDL2017 neutron activation database to find all isotopes/isomers that have both TENDL and PENDF files, and asynchronously downloads them. While this is not explicitly necessary for the use of `preprocess_fendl3.py`, at this point, I do not know of a better way for the user to download the entirety of the relevant/usable FENDL3 isotopes directly from TENDL2017, without having to do so manually. @gonuke if you think this would be valuable here, I can include it too.

tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -105,8 +105,9 @@ def extract_gendf_pkza(gendf_path):
     Z, element, A = first_line.split('-')[:3]
 
     Z = int(Z)
+    isomer_tag = 'm' if 'm' in A.lower() else 'n'
     M = 2 if 'n' in A.lower() else 1 if 'm' in A.lower() else 0
```
eitan-weinstein, line 109: Per the conversation in the fusion-activation channel on Slack, I'm pretty sure that the 'n' tag corresponds to a higher energy metastable state. Just flagging this here, though, in case we find that it represents something else.

PR #98, opened 2025-09-17
Building ALARAJOY into ALARA DataLib

### LINKED ISSUE #79: Extend ALARA's ASCII data library reading capability to read new data
ALARA's data handling has some infrastructure for writing new classes that will read ASCII based nuclear data.  All of the existing cases are based on legacy data formats, often designed to be robust for many cases.  With new data handling capability, we have the ability to write activation data that does not need to be robust to many cases, but can be simple.  This new data format will be some version of multiple entries were each entry is:
* parent nuclide
* child nuclide
* G group cross-sections, where G is the number of groups

The choice of delimiter and/or line wrapping should be based on what ever choices simplify the C++ code development since the data will be written to that format by python that tends to be more flexible.

This task will be complete when ALARA can read a data file of some newly defined data format.

### LINKED ISSUE #80: Modify data processing tools to write data in correct format
Once a suitable format has been determined (#79), modify the tools that process new data to write the data in the appropriate format.  (This task may be moot if the selected format matches what these tools already do.)

eitan-weinstein:
This PR introduces a new C++ file/header pair into `src/DataLib` called `ALARAJOY.cpp` and `ALARAJOY.h`. Similarly to other data libraries in `src/DataLib`, ALARAJOY is designed to be callable like EAFLib, IEAFLib, etc. through the `convert_lib` input block in the ALARA input file. Unlike EAFLib or IEAFLib, however, ALARAJOY is not operating on a specific data library, per se, but rather on a preprocessed CSV format of FENDL3 (TENDL2017) data, which must first be converted to the necessary groupwise format in the ALARAJOY Python wrapper found at `tools/ALARAJOYWrapper/preprocess_fendl3.py`. Despite this difference, usage of ALARAJOY in the `convert_lib` input largely follows the EAFLib format, with the general format of the `convert_lib` block for ALARAJOY being as such:

`convert_lib ajoylib alaralib transFname decayFname alaraFname`

wherein:

- `ajoylib` is the new library value for ALARAJOY
- `transFname` is the file path to the CSV containing the preprocessed FENDL3 transmutation data
- `decayFname` is the file path to EAF decay data.
- `alaraFname` is the file path template for resultant ALARA data libraries (.lib, .idx. .gam, .gdx)

It should be noted that ALARAJOY is a hybrid method, in that it is sourcing transmutation data from FENDL3 and decay data from EAF. As a result, this PR also includes a number of changes to `EAFLib.C` and `EAFLib.h` to create a decay-only constructor that can be accessed from ALARAJOY, as well as a number of public decay related functions in the class. Testing is still underway to ensure that the decay data is being properly handled, as compared to pure EAF data library conversions.

Closes #80 
Closes #79


src/DataLib/ALARAJOY.cpp
```diff
@@ -0,0 +1,181 @@
...
+        size_t a = item.find_first_not_of(" \t");
+        size_t b = item.find_last_not_of(" \t");
+        if (a == std::string::npos) continue;
+        xs.push_back(std::stof(item.substr(a, b - a + 1)));
+    }
+    return xs;
+}
+
+// Free helper: Parse a single CSV row
+static CSVRow parseCSVRow(const std::string& line)
```
gonuke, line 32: Now that I look at this, it seems like it might be easer to have space-delimited files rather than comma-delimited.  That should let you use simple ">>" operators to get the data you want.

You'd end up doing something like:
```
ss >> index >> row.parentKza >> row.daughterKZA >> row.emittedParticles >> row.nonZeroGroup;
row.crossSections = std::vector<float>(row.nonZeroGroup);
for (int g=0; g < row.nonZeroGruop; g++) { 
   ss >> row.crossSection[g];
}
```

gonuke, line 32: This should be a member function of the ALARAJOY class.  (or maybe CSVRow is a class and this is a member function of that?)

src/DataLib/ALARAJOY.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include <vector>
+#include <string>
+#include <cstring>
+#include <cstdio>
+
+// Define static member
+std::vector<CSVRow> ALARAJOYLib::csvData;
+
+// Free helper: Parse cross section array from Python list (of floats) format
+static std::vector<float> parseXSectionArray(const std::string& arrayStr) {
```
gonuke, line 14: This should be a member function of the ALARAJOY class. (if it's needed)

src/DataLib/ALARAJOY.cpp
```diff
@@ -0,0 +1,181 @@
...
+    if (!csvData.empty()) {
+        currentParent = csvData[0].parentKZA;
+    }
+}
+
+// Read library header information
+void ALARAJOYLib::getTransInfo()
+{
+    // Fixed for Vitamin J energy group structure
+    nGroups = 75;
```
gonuke, line 124: This should be 175... and not sure we want this to be hard coded?

src/DataLib/ALARAJOY.cpp
```diff
@@ -0,0 +1,181 @@
...
+    if (currentRowIndex >= csvData.size()) {
+        return LASTISO; // end of data
+    }
+
+    int rxnNum = 0;
+    int parentKZA = currentParent;
+
+    // Read all reactions for current parent
+    while (currentRowIndex < csvData.size() && 
+    csvData[currentRowIndex].parentKZA == currentParent)
```
gonuke, line 151:
```suggestion
           csvData[currentRowIndex].parentKZA == currentParent)
```

src/DataLib/ALARAJOY.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+        // Store reaction data
+        transKza[rxnNum] = row.daughterKZA;
+        strcpy(emitted[rxnNum], row.emittedParticles.c_str());
+
+        // Fill cross sections (75 groups total)
+        for (int g = 0; g < nGroups; g++)
+        {
+            xSection[rxnNum][g] = 
+            (g < row.nonZeroGroups) ? row.crossSections[g] : 0.0;
```
gonuke, line 163:
```suggestion
                (g < row.nonZeroGroups) ? row.crossSections[g] : 0.0;
```

src/DataLib/ALARAJOY.cpp
```diff
@@ -0,0 +1,181 @@
...
+    while (currentRowIndex < csvData.size() && 
+    csvData[currentRowIndex].parentKZA == currentParent)
+    {
+        const CSVRow& row = csvData[currentRowIndex];
+
+        // Store reaction data
+        transKza[rxnNum] = row.daughterKZA;
+        strcpy(emitted[rxnNum], row.emittedParticles.c_str());
+
+        // Fill cross sections (75 groups total)
```
gonuke, line 159: Should be 175 groups, but we may want to this to be flexible

gonuke (changes requested):
This is coming together nicely - especially with this new inheritance.

tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -161,16 +161,25 @@ def iterate_MTs(MTs, file_obj, mt_dict, pKZA):
...
-            }
+            }  
         )
 
+        if len(sigma_list) > max_groups:
+            max_groups = len(sigma_list)
+
+    for row, sigma_list in zip(gendf_data, sigma_lists):
+        for i in range(max_groups):
+            row[f'XS {i}'] = sigma_list[i] if i < len(sigma_list) else None
```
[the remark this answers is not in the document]
gonuke: It seems like this may be designed as a way to write all the data using pandas CSV writer...

I wonder if it's cleaner to just write it out with a bespoke method?

tools/ALARAJOYWrapper/mt_table.csv
```diff
@@ -12,25 +12,25 @@ MT,Reaction,Description
 19,"(n,f)",First-chance neutron-induced fission
 20,"(n,nf)",Second-chance neutron-induced fission
 21,"(n,2nf)",Third-chance neutron-induced fission
-22,"(z,nα)","Production of a neutron and an alpha particle, plus a residual"
```
gonuke, line 15: Redundant

tools/ALARAJOYWrapper/files_for_tests/mt_test2.csv
```diff
@@ -1,6 +1,6 @@
 MT,Reaction,Description
 1,"(n,total)",Neutron total cross sections.
-30,"(z,2n2α)","Production of two neutrons and two alpha particles, plus a residual."
```
gonuke, line 3: Redundant

src/DataLib/EAFLib.h
```diff
@@ -112,17 +112,20 @@ class EAFLib : public ASCIILib
...
-  int getGammaData();
+
+public:
 
   /* Interface from ASCIILib */
   void getTransInfo();
   void getDecayInfo();
   int getTransData();
   int getDecayData();
+  int getGammaData();
```
gonuke, line 123: With inheritance, inheritance don't think any of these need to be public

src/DataLib/ALARAJOY.cpp
```diff
@@ -0,0 +1,149 @@
...
+    std::getline(inTrans, line);
+
+    // Read all data rows
+    while (std::getline(inTrans, line))
+    {
+        if (line.empty()) continue; // Skip empty lines
+
+        DSVRow row = parseDSVRow(line);
+        dsvData.push_back(row);
+    }
```
gonuke, line 65: If we are confident enough in the formatting, we should be able to just read directly from `inTrans` rather than reading lines one at a time and parsing them.  Maybe we need to discuss...

src/DataLib/ALARAJOY.cpp
```diff
@@ -0,0 +1,149 @@
...
+    nParents = 0;
+
+    // Allocate arrays for maximum reactions
+    transKza = new int[MAXALARAJOYRXNS];
+    xSection = new float*[MAXALARAJOYRXNS];
+    emitted = new char*[MAXALARAJOYRXNS];
+
+    for (int rxnNum = 0; rxnNum < MAXALARAJOYRXNS; rxnNum++) {
+        xSection[rxnNum] = nullptr;
+        emitted[rxnNum]  = new char[6];
```
gonuke, line 86: magic number: can we define the maximum length of an emitted string somewhere and use that?

PR #100, opened 2025-09-17
Addressing ALARA memory issues as they related to ALARAJOY

eitan-weinstein:
I've come across a number of areas with variable deletion that caused memory issues when I've tried to implement the new changes for ALARAJOY. I'm interested to compare these with #92 to find overlap in issues identified and implement a robust fix for ALARA's memory handlng.


eitan-weinstein:
> I wonder if we should merge this because it improves things and continue hunting for more improvements in a different PR? Thoughts?

Yeah, I think that's sensible. These improvements definitely cover most of the issues that I had been running into for my day-to-day ALARA usage and inevitably we're going to need to continue patching going forward.

gonuke (approved):
This is a good step forward...

PR #101, opened 2025-09-18
Handling Arbitrary Isometric States in ALARAJOY.

### LINKED ISSUE #95: Handle arbitrary isomeric states in ALARAJOY
We currently seek and handle isomeric states indicated with 'm' and 'n', but not higher.  We should consider how to make this more general and/or issue an error/warning when a higher states is encountered.

eitan-weinstein:
This PR resolves #95 by creating a truncated alphabet dictionary within the `extract_gendf_pkza()` function in `tendl_processing.py` ranging from m-z. This dictionary corresponds to the starting count of isomeric states in TENDL where m = 1 corresponds to the first excitation level. However, we would only really expect values of 'm' or 'n', corresponding to the first two isomeric states in TENDL2017, so even as the function can handle higher states, will raise a warning for higher states.


gonuke (approved):
This looks pretty clean... but I realize we don't have any unit test for any of this.  We'll have to make an issue for that.

Issue #105, opened 2025-09-18
Assess and determine a database technology


gonuke:
For the specific database design we are considering, what is a good DB technology?

Considerations:
* total data size
* API for slicing and selecting data
   * usability
   * efficiency

gonuke:
Start with Pandas and some of their standard file options and see how it performs.

gonuke:
Expected database table design:
```
* parent nuclide
* normalized spectrum type  -----
* total irradiation time.       |--- DB index
* average flux magnitude.   -----
* child nuclide
* number density
```

anu1217:
Here's a high-level comparison/overview of what I've found out about hdf5 and SQLite specifically.

| Attribute | hdf5 | SQLite |
|:----------:|:--------:|:---------:|
| Ease of setup | Requires dedicated software/packages to read and interpret data (e.g. h5py) due to binary format | Serverless & stores data in a single file (can be used in Python through sqlite3) |
| Ease of usage | Data stored in filesystem-like format | Relational database that can hold data in multiple 2D tables |
| Ease of usage | Verbose syntax due to large number of variables created | Only takes one user connection at a time to a database |
| Data storage | Consistently performs well with large datasets and quickly saves and retrieves data | Slower to query large datasets and create an overall representation of data |
| Data slicing | Can be accessed like NumPy arrays, manipulated using numpy attributes | Data queried using verbose SELECT & filter statements (e.g. WHERE) |
| Data stability | Potential data corruption issues over time | Backwards compatible, cross-platform stable |
| Data stability | Lack of fluidity & large performance overhead for frequently updated files | Datatypes are advisory, not mandatory - can lead to cross-platform consistency issues |

I think the relevance of some of these considerations depends on whether we're adding data or querying existing data more frequently. In the latter case I can see hdf5 being the more suitable format.

gonuke:
Let's go with SQLite under Pandas and try it, with the option translate everything to HDF5 if we find a need in the future.

Issue #89, opened 2025-09-06
ALARA data formatting/conversion for use with output comparison tools


gonuke:
The output format of ALARA data may requires some parsing/processing before it can be used in comparison tools.  These can either be methods that process data on the fly, or  standalone scripts that translate the data from the native format to some intermediate file format for repeated use in comparison.

PR #114, opened 2025-09-24
Add tool to write ALARA output from mesh-based input to pandas dataframe

anu1217:
This script takes an ALARA output file (and corresponding input, material library, and flux files) and writes the parent element, irradiation time, normalized flux, daughter nuclides, and number densities to a pandas dataframe.

Some assumptions and limitations of this code at the time of this PR:
1) Works with an ALARA flux that is discretized over a MOAB mesh object
2) Takes 1 single material & element over the mesh
3) Takes 1 single irradiation time in the 'schedule' input block
4) Relies on pyne.alara.num_density_to_mesh(), which does not work correctly if the output file produces nuclides in the m2/n state
5) The dataframe will also eventually need a time-dependent flux magnitude

Addresses #89 & #105


anu1217:
The script now allows the user to choose between processing mesh-based and meshless output. For the mesh-based case, each mesh voxel can take 1 separate material/element. The irradiation schedule still takes only 1 entry.

PR #116, opened 2025-09-25
Building ALARAJOY into ALARA DataLib (Clean)

### LINKED ISSUE #79: Extend ALARA's ASCII data library reading capability to read new data

### LINKED ISSUE #80: Modify data processing tools to write data in correct format

eitan-weinstein:
Closes #98 .

I made a mistake that I couldn't figure out how to resolve naturally when rebasing #98 , so I moved the new changes to a new branch to start fresh with a clean PR.

Quoted below is the previous PR description:

> This PR introduces a new C++ file/header pair into `src/DataLib` called `ALARAJOY.cpp` and `ALARAJOY.h`. Similarly to other data libraries in `src/DataLib`, ALARAJOY is designed to be callable like EAFLib, IEAFLib, etc. through the `convert_lib` input block in the ALARA input file. Unlike EAFLib or IEAFLib, however, ALARAJOY is not operating on a specific data library, per se, but rather on a preprocessed CSV format of FENDL3 (TENDL2017) data, which must first be converted to the necessary groupwise format in the ALARAJOY Python wrapper found at `tools/ALARAJOYWrapper/preprocess_fendl3.py`. Despite this difference, usage of ALARAJOY in the `convert_lib` input largely follows the EAFLib format, with the general format of the `convert_lib` block for ALARAJOY being as such:
> 
> `convert_lib ajoylib alaralib transFname decayFname alaraFname`
> 
> wherein:
> 
> * `ajoylib` is the new library value for ALARAJOY
> * `transFname` is the file path to the CSV containing the preprocessed FENDL3 transmutation data
> * `decayFname` is the file path to EAF decay data.
> * `alaraFname` is the file path template for resultant ALARA data libraries (.lib, .idx. .gam, .gdx)
> 
> It should be noted that ALARAJOY is a hybrid method, in that it is sourcing transmutation data from FENDL3 and decay data from EAF. As a result, this PR also includes a number of changes to `EAFLib.C` and `EAFLib.h` to create a decay-only constructor that can be accessed from ALARAJOY, as well as a number of public decay related functions in the class. Testing is still underway to ensure that the decay data is being properly handled, as compared to pure EAF data library conversions.
> 
> Closes #80 Closes #79


gonuke (changes requested):
This is all getting really close... thanks @eitan-weinstein

src/DataLib/ALARAJOY.cpp
```diff
@@ -0,0 +1,134 @@
...
+        dsvData[currentRowIndex].parentKZA == currentParent)
+    {
+        const DSVRow& row = dsvData[currentRowIndex];
+
+        // Store reaction data
+        transKza[rxnNum] = row.daughterKZA;
+        strcpy(emitted[rxnNum], row.emittedParticles.c_str());
+
+        // Fill non-zero cross sections
+        int count = std::min(row.nonZeroGroups, nGroups);
```
gonuke, line 113: It doesn't seem logical that `nonZeroGroups` can ever be larger than `nGroups`, given the way they are both defined?

gonuke (changes requested):
Thanks for the latest changes... we can chat about the sizing of the ultimate cross-section vector soon.

gonuke (changes requested):
A few final whitespace and simplifications

eitan-weinstein:
By reorganizing the writing of the DSV in `preprocess_fendl3.py` to be written in ascending order of parent KZA, the ALARAJOY-processed idx file is looking much better, no longer having separated parents, and seeming to include the singular, correct half-life (from having spot-checked a few parents). There remains more to the process, still, as now ALARA stalls out in a loop of node checking of the form:

>```
>                        Checking node 11820
>                                Set truncation state: 0 (-nan)
>                        Processing new node 220460 at rank 2332
>                                Looking for data for 220460
>                                        Found data for 220460 at offset 677825
>                                        Read 24 reaction path(s) for 220460.
>                                Setting NuclearData members.
>                                Running truncation reference calculation.
>```

I let it iterate through nodes and ranks for a while, but it became clear that it would not resolve. The source is less obvious from the idx now that it seems to be formatted and written correctly, so I'm going to start digging through `Chain.C`, where this output originates.

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -14,6 +13,49 @@ def args():
...
+    Returns:
+        None 
+    """
+
+    xs_key = 'Cross Sections'
+    join_keys = list(cumulative_data[0].keys())
+    join_keys.remove(xs_key)
+    
+    # Sort list of reaction dictionaries by ascending parent KZAs
+    cumulative_data.sort(key=lambda rxn: rxn.get(join_keys[0], float('inf')))
```
gonuke: My assumption is that it was already sorted into blocks of the same parent KZA, but those parent KZAs are ordered based on the alphabetical ordering of the element symbols?

gonuke (approved):
This seems to have reached a point where it is ready for more validation after merging the main improvements.  Thanks @eitan-weinstein for working through all of this.

Issue #115, opened 2025-09-25
Create an ALARA input file with other necessary input for a single element


gonuke:
Create a typical ALARA input file (with other input files, e.g. fluxin) for a simple irradiation of a single element, (e.g. Fe), at a typical flux for a typical steady duration (e.g. 20 years).

gonuke:
PR these files to an `examples` directory in this repo

PR #122, opened 2025-09-30
Add script to produce ALARA material library for each individual element

anu1217:
Writing a script that will write a material library for all elements. I anticipate that this will be useful for populating the database of irradiation histories, and for addressing issues such as #115.


gonuke:
It's not clear to me what problem this is solving since it is already possible to define an ALARA `mixture` as being a single `element` without having to define a `material` first

PR #123, opened 2025-09-30
Add example input file of Fe irradiated for 20 years

### LINKED ISSUE #115: Create an ALARA input file with other necessary input for a single element

anu1217:
This is an ALARA input file of Fe irradiated for 20 years in the FNSF FW spectrum. Custom filepaths for the standard ALARA library and element library are used.

fixes #115

### DIFF
--- a/examples/singleElement_20y_inp/alara_inp_fe_20y
+++ b/examples/singleElement_20y_inp/alara_inp_fe_20y
@@ -35,5 +35,4 @@ end
 
 #other parameters
 truncation 1e-3
-impurity 5e-6 1e-3
 dump_file dump_1w.file


examples/alara_inp_fe_20y
```diff
@@ -0,0 +1,38 @@
...
+material_lib ../data/elelib_to_matlib.txt
+element_lib /root/data/elelib.std
+data_library alaralib /opt/groupspace/shared/n/nukecode/ALARA/data/fendl2bin
+
+output zone
+        number_density
+end
+
+#     flux name    fluxin file   norm   shift   unused
+flux  my_flux       fluxfnsfIBfw.txt 1e18     0      default
```
gonuke, line 24: The flux file has some groups with fluxes up to 1E13 or higher.  Are we sure we want to multiply by 1E18 here?

What fusion power is this ?

anu1217: The 1e18 was leftover from a previous file, but looking at an FNSF neutronics paper, a fusion power of 518MW is assumed for neutronics studies. So 518 MW / 17.6 MeV ~ 1.8E20 reactions/s or n/s. Or should we be looking at normalization differently?

anu1217: I also see that there is some normalization info provided in the input file that goes with the flux - will check in with Tim on this

examples/singleElement_20y_inp/alara_inp_fe_20y
```diff
@@ -0,0 +1,38 @@
...
+
+mat_loading
+    zone_1    mix_1
+end
+
+mixture mix_1
+    material mat:fe 1 1.0
+end
+
+material_lib fe_matlib.txt
```
gonuke: I'm ok with the introduction of a custom material library, but probably makes more sense for something that isn't just defining a raw element

examples/singleElement_20y_inp/alara_inp_fe_20y
```diff
@@ -0,0 +1,38 @@
...
+element_lib /root/data/elelib.std
+data_library alaralib /opt/groupspace/shared/n/nukecode/ALARA/data/fendl2bin
```
gonuke: We may want to give readers instructions to have a copy or symlink to this library in the current directory, and let them figure out some of the details.  In particular, they won't have access to this path.
```suggestion
data_library alaralib fendl2bin
```

anu1217: The fendl2bin library isn't readily available for non-CAE users to find, correct? If that's the case I've added a comment at the top of the file about using the convert_lib option to make the library

examples/fluxfnsfIBfw_518MW.txt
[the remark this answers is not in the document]
anu1217: Actually, this flux is the same as the one Tim used for his example - we could keep one and delete the other.

I think it makes sense to add a subdirectory for the flux files if we anticipate adding more examples without necessarily making a new flux file for each one.

gonuke (approved):
LGTM - thanks @anu1217

PR #124, opened 2025-09-30
Create initial examples for practical usage of ALARA

bohmt:
Description
Add an examples directory filled with practical usage examples and associated input files

Motivation and Context
This change provides practical usage examples and documentation for new users.
This is intended to fix issue [#115]

Changes
This provides documentation updates.

Behavior
The current repo does not provide practical usage examples.  The update provides some practical examples.

Other Information
None


bohmt:
Oops, I missed that anu1217 had a similar PR [#123]

examples/Results_singleElement_Fe_FENDL2/singleElement.ala
```diff
@@ -0,0 +1,81 @@
...
+		0.0
+        1       5.0
+end
+
+mat_loading
+        inner_zone1  mix1
+end
+
+material_lib  sampleMatlib
+element_lib  nuclib
```
anu1217, line 30: Maybe we can add an example filepath here (especially since there is an element library built into the ALARA data folder)?
```suggestion
element_lib  ALARA/data/elelib.std
```

examples/Results_singleElement_Fe_FENDL2/singleElement.ala
```diff
@@ -0,0 +1,81 @@
...
+output interval
+        units Bq kg
+        number_density
+        specific_activity
+	total_heat
+	dose contact fendl2bin ANS6_4_3
+end
+## 
+truncation  1e-11
+
```
anu1217, line 81: Can we add an example of an impurity tolerance here?

bohmt: I don't think impurities will be relevant in this particular example because the intent is for a single element rather than a mixture/alloy.  As we create additional examples, it might be a good idea to include the impurity tolerance in those input files.

Issue #120, opened 2025-09-29
Spot check some of nuclides vs TENDL's published PENDF files


gonuke:
For a small selection of nuclides, compare the PENDF files that we generate to the PENDF files that are published.

If there are any differences, we need to understand why and decide which is correct.

PR #125, opened 2025-09-30
Expanding NJOY Usage to Create PENDF Files as part of NJOY Run

### LINKED ISSUE #118: Review NJOY input files for various nuclides to identify similarities and differences
This issue will be resolved by having a more complete picture of which NJOY modules we should be invoking.  Almost certainly:
* RECONR
* BROADR
* UNRESR
* GROUPR

Perhaps also:
* GASPR

Probably not necessary for activation:
* THERMR
* HEATR
* PURR

### LINKED ISSUE #119: Expand calls to NJOY to invoke a full range of modules
Update NJOY use to invoke all the necessary modules directly on the TENDL data:
* RECONR
* BROADR
* UNRESR
* GASPR (?)
* GROUPR

eitan-weinstein:
Closes #118 
Closes #119

This PR modifies the ALARAJOYWrapper to use NJOY more robustly by creating its PENDF files in house, instead of relying on their availability from the TENDL2017 data base. There are a number of key isotopes in  TENDL 2017 that only have a TENDL file available, rather than a preprocessed PENDF as well. In order to convert to the groupwise cross-section data needed for ALARA, NJOY's GROUPR module uses both a TENDL/ENDF and a PENDF file pair to perform this conversion. However, TENDL 2017 provides the NJOY input files uses to produce said pointwise data, so by following the general format found for isotopes that do have available PENDFs, I expanded the `njoy_input` `Template` to accommodate for that.

As an initial test, I have run this on <sup>1</sup>H and <sup>2</sup>H to confirm that isotopes that otherwise only have a TENDL file in the database can be converted and produce reasonable (although yet unvalidated) results. Moreover, I have tested this on <sup>56</sup>Fe to start investigating and begin the verification process needed for #120 .

### DIFF
--- a/tools/ALARAJOYWrapper/njoy_tools.py
+++ b/tools/ALARAJOYWrapper/njoy_tools.py
@@ -12,19 +12,19 @@ def set_directory():
     Arguments:
         None
     Returns:
-        dir (str): Path to the current working directory (CWD) from which the
-            command was called.
+        dir (PosixPath): Path to the current working directory (CWD) from
+            which the command was called.
     '''
 
-    dir = str(Path.cwd())
-    fendl_dir = 'ALARAJOYWrapper'
-    if fendl_dir not in dir:
-        dir += f'/{fendl_dir}' # Should only be called from one directory up
+    dir = Path.cwd()
+    fendl_dir = Path('ALARAJOYWrapper')
+    if not (dir.parents[0] / fendl_dir).exists():
+        dir = dir / fendl_dir # Should only be called from one directory up
     return dir
 
 # Define constant(s)
 dir = set_directory()
-INPUT = dir + '/njoy.inp'
+INPUT = dir / Path('njoy.inp')
 
 # Create input file general template
 njoy_input = Template(Template(
@@ -216,7 +216,7 @@ def ensure_gendf_markers(gendf_path, matb):
         The formatting for these records can be found at:
         https://t2.lanl.gov/nis/endf/intro06.html
     Arguments:
-        gendf_path (str): File path to the newly created GENDF file.
+        gendf_path (PosixPath): File path to the newly created GENDF file.
         matb (int): Unique material ID for the material in the GENDF file.
     
     Returns:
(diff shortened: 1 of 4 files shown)


gonuke:
I think for now we should only allow 9 isomeric states and issue a warning for the rest.  The KZA formula is pretty deeply embedded throughout ALARA and it's probably a big update to allow for this new format.

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -245,14 +293,22 @@ def run_njoy(element, A, matb):
...
-        ensure_gendf_markers(gendf_path, matb)
+        for filetype in save_dict.values():
+            save_path = f'{dir}/{
+                        filetype['Specific Dir']
+                        }/tendl_2017_{element}{str(A).zfill(3)}'
+            
+            if not os.path.isdir(
+                os.path.join(dir, filetype['Specific Dir'])
+                ):
+                    os.mkdir(os.path.join(dir, filetype['Specific Dir']))
```
gonuke, line 304: `Path.mkdir()` allows you to set `exist_ok=True` so you can just make the directory even if it exists already.

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -230,13 +266,25 @@ def run_njoy(element, A, matb):
...
+        ('PENDF', 'pendf_files', '.pendf', 25)
+    ]
+    save_dict = {
+        name: {
+            'Specific Dir'              :           spec_dir,
+            'Extension'                 :                ext,
+            'Tape'                      :               tape
+        }
+        for name, spec_dir, ext, tape in file_metadata
+    }
```
gonuke: I wonder if it's simpler to just build the dictionary from scratch rather than expanding the tuples into a dictionary:
```
file_metadata = {
    'GENDF' : { 'Specific Dir' : 'gendf_files', 'Extension' : '.gendf', 'Tape' : 31 }
    'PENDF' : { 'Specific Dir' : 'pendf_files', 'Extension' : '.pendf', 'Tape' : 25 }
}
```
I might even advocate for shorter keys in the nested dictionaries: `dir', `ext`, `tape`

gonuke (changes requested):
I think it would be good to comb over everything to see how using `Path` objects throughout instead of strings that represent directories/filenames/paths might allow for simpler operations.

I see a few places where you are converting strings to `Path` but maybe they could have just been a `Path` to start with?

I guess this is a little tangential to the purpose of this PR, but a good opportunity to clean things up in this respect

gonuke:
This appears to need a rebase?

PR #126, opened 2025-10-06
Allowing ALARAJOYWrapper to be called from anywhere.

### LINKED ISSUE #96: Allow ALARAJOY wrappers to be called from anywhere
We currently restrict which directories are valid for invoking ALARAJOY wrappers.  We should consider how to be more relaxed and robust to this as users may want to call these from various places in their filesystem

eitan-weinstein:
Closes #96 .

Modifies `set_directory()` in `groupr_tools.py` to locate the directory `ALARAJOYWrapper` from anywhere, so that the user can call the preprocessor from anywhere.

### DIFF
--- a/tools/ALARAJOYWrapper/groupr_tools.py
+++ b/tools/ALARAJOYWrapper/groupr_tools.py
@@ -13,11 +13,11 @@ def set_directory():
     Arguments:
         None
     Returns:
-        dir (str): Path to the current working directory (CWD) from which the
-            command was called.
+        dir (pathlib._local.PosixPath): Path to the current working directory
+            (CWD) from which the command was called.
     '''
 
-    return str(Path(__file__).resolve().parent)
+    return Path(__file__).resolve().parent
 
 # Define constant(s)
 dir = set_directory()


gonuke:
Perhaps first I need to understand all the directories that matter here.  It seems to me that there are three relevant directories:
1. the directory that contains the python files so that they can be imported correctly
2. the directory that contains the original ENDF Files
3. the directory where the temporary processing files will be generated and the ultimately DSV file created

I am most concerned about item 1 because it seems that if someone invokes the `preprocess_fendl3.py` command from a directory that does not contain the other python scripts, it will fail to find them on import.

In this change, you expand on how you set `dir` by searching the file tree, but it seems like that is used primarily for item 3 above.

gonuke:
I think this PR is working much harder than I was imagining....  Perhaps I need to better understand how you normally invoke this now.  I'm imagining a future in which someone has a bunch of data downloaded in directory `/root/some/path/to/data/tendl-2017` and they go to that directory
```
cd /root/some/path/to/data/tendl-2017
```
and then invoke this script
```
python3 /root/another/path/to/processing_script.py
```
In order for that to work, the script will need to know how to import the modules you have written.

Most of what I am looking for here is someway to make sure the top-level script can find the modules by updating `sys.pythonpath` or something like that.

PR #127, opened 2025-10-09
Add markdown document to summarize fusion parameters for database

### LINKED ISSUE #86: Determine the range of irradiation histories to span the database
An irradiation history is described, in general, but the neutron flux as a function of energy and time, 
$\phi(E,t)$.

In general, this can be separated into a flux spectrum, $f(E)$, and a time varying magnitude, $\hat\phi(t)$, such that $\phi(E,t) = f(E)\hat\phi(t)$.

If the duration of the irradiation history is $t \in [0,t_{irr}]$, then one approximation of the time-varying magnitude is $\bar\phi = \frac{1}{t_{irr}}\int_0^{t_{irr}} \hat\phi(t)dt$.

This results in three quantities that can be used to define an irradiation history:
* $f(E)$ from discrete set of flux spectrum shapes
* $\bar\phi$ the average flux magnitude over the irradiation history
* $t_{irr}$ the total time of the irradiation history

For each of these dimensions, a range of interest should be defined over which a set of activation calculations will be performed.

anu1217:
This document provides a summary of quantities required for a fusion activation database, and provides some examples of typical values.

fixes #86

### DIFF
--- a/examples/fusion_irr_hist.md
+++ b/examples/fusion_irr_hist.md
@@ -1,6 +1,6 @@
 # Irradiation History-Related Parameters for Activation Database
 
-## This document summarizes the quantities of interest for a fusion materials activation database. An underlying idea for describing the irradiation history is that neutron flux history $\phi(E,t) = f(E)\phi\hat(t)$, where $f(E)$ is a normalized flux spectrum shape, and $\phi\hat(t)$ is the respective magnitude.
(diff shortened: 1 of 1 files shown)


examples/fusion_irr_hist.md
```diff
@@ -0,0 +1,22 @@
+# Irradiation History-Related Parameters for Activation Database
+
+## This document summarizes the quantities of interest for a fusion materials activation database. An underlying idea for describing the irradiation history is that neutron flux history $\phi(E,t) = f(E)\phi\hat(t)$, where $f(E)$ is a normalized flux spectrum shape, and $\phi\hat(t)$ is the respective magnitude.
```
gonuke: I don't think this should be a heading, should it?

gonuke (approved):
Thanks @anu1217

PR #113, opened 2025-09-22
Pandas postprocessor python script

eitan-weinstein:
This PR responds to #89 and follows the output methods and formatting in #111 . Currently, this PR includes a singular Python script that can be called to read through a piped ALARA output file, identify all tabular data, and store each individual table into a Pandas DataFrame to be saved as separate CSV files.

PR #135, opened 2025-10-22
Comparison Tool for Single Initial Nuclide Response

### LINKED ISSUE #84: Create a tool to compare the results of a single response for a single initial nuclide
A script and/or python notebook will take two dataset that represent a single activation response (i.e. number density, specific heat, total photon source, etc) for a single initial nuclide, and then perform summary operations and generate plots to facilitate comparison.

eitan-weinstein:
Closes #84 .

I've included a Jupyter notebook and an accompanying Python module for the comparison of ALARA output data from an ALARA run based on the input scripts and data from `examples/singleElement.ala`. This Jupyter notebook allows for the comparison between the standard fendl2 data in `fendl2bin`, as well as ALARAJOY-processed data from fendl3. I converted the data from both output files to CSV files, following the meshless table extraction method in #114. 

I'm still working on expanding and building all of my comparison tools, so I'm only uploading this as a draft for now, but I wanted to put up my work that I have thus far ahead of tomorrow's fusion-activation meeting.

The major first impression from the plots that I am seeing is that ALARAJOY is not allowing for proper tritium tracking, as we know both from fendl2 and FISPACT that tritium is highly significant on the century timescale for specific activity and decay heat in particular, but the ALARAJOY-processed fendl3 data is not showing any tritium. 

748053bc-3475-4aa1-861d-0924eac1aa71

More analysis and plots to come, but I wanted to get this up so that we can start discussing initial results and next steps in debugging ALARAJOY's tritium issue.


tools/single_nuc.py
```diff
@@ -0,0 +1,669 @@
...
+        inp_datalibs=[],
+        inp_variables=[],
+        inp_units=[]):
+    
+    datalibs = ['fendl2', 'fendl3']
+    variables = [
+        'Number_Density', 'Specific_Activity',
+        'Total_Decay_Heat', 'Contact_Dose'
+        ]
+    units = ['atoms_kg', 'Bq_kg', 'W_kg', 'Sv_hr_._data_ANS6_4_3']
```
gonuke: These should probably be bound together in some kind of dictionary.  However, you might want to note that ALARA is flexible and allows different units for these same quantities.  We may need to handle this differently....

tools/single_nuc.py
```diff
@@ -0,0 +1,669 @@
...
+        sublist (list): Intersection of x and y.
+        indices (list): Index or indices of intersection on x.
+    '''
+
+
+    sublist = [
+        val for i, val in enumerate(x, start=1)
+        if val in y or i in (y if isinstance(y, list) else [y])
+        ]
+    indices = [i for i, val in enumerate(x) if val in sublist]
```
gonuke: It might be simpler to use `list.index()` for this: https://stackoverflow.com/questions/176918/how-can-i-find-the-index-for-a-given-item-in-a-list

tools/single_nuc.py
```diff
@@ -0,0 +1,669 @@
...
+    Returns:
+        sublist (list): Intersection of x and y.
+        indices (list): Index or indices of intersection on x.
+    '''
+
+
+    sublist = [
+        val for i, val in enumerate(x, start=1)
+        if val in y or i in (y if isinstance(y, list) else [y])
+        ]
```
gonuke: What if you convert them both to sets and use set operations?
```
sublist =  list(set(x).intersection(set(y)))
```

tools/single_nuc.py
```diff
@@ -0,0 +1,669 @@
...
+    
+    datalibs = ['fendl2', 'fendl3']
+    variables = [
+        'Number_Density', 'Specific_Activity',
+        'Total_Decay_Heat', 'Contact_Dose'
+        ]
+    units = ['atoms_kg', 'Bq_kg', 'W_kg', 'Sv_hr_._data_ANS6_4_3']
+
+    if inp_datalibs:
+        datalibs, _ = select_sublist(datalibs, inp_datalibs)
```
gonuke: You might not need the `select_sublist()` method if these are replaced by a single line if using set intersection (see above)

tools/single_nuc.py
```diff
@@ -0,0 +1,669 @@
...
+
+    sublist = [
+        val for i, val in enumerate(x, start=1)
+        if val in y or i in (y if isinstance(y, list) else [y])
+        ]
+    indices = [i for i, val in enumerate(x) if val in sublist]
+    return sublist, indices
+
+def process_metadata(
+        gen_csv='',
```
gonuke: Instead of reading CSV files, why not import your script from #113 and call `parse_tables()` to get the data straight into dataframes?

gonuke (changes requested):
A bunch of comments on all the code to read certain CSV files, perhaps superseded by the suggestion to simply use the `parse_tables` function from the script in #113

gonuke (changes requested):
These tools seemed to be aimed at the QA of the library processing, so I wonder if they should live in a subdirectory of `ALARAJOYWrapper`?

Also, I'd love to review this in pieces.  You appear to have a progression of capabilities here:
1. setup and run a new simulation
2. load the results of one or more simulations into a data structure
3. a variety of independent post-processing and plotting steps - each of which can be added/reviewed separately

Could you break this PR up into some of these steps, please?

PR #136, opened 2025-10-24
ALARAJOY Gas Production

eitan-weinstein:
PR to update the Python elements of the ALARAJOYWrapper to be able to handle MTs 203-207 (gas production totals).

### DIFF
--- a/tools/ALARAJOYWrapper/preprocess_fendl3.py
+++ b/tools/ALARAJOYWrapper/preprocess_fendl3.py
@@ -5,8 +5,6 @@ import njoy_tools as njt
 import argparse
 import warnings
 from pathlib import Path
-from pandas import DataFrame
-from numpy import array
 
 def args():
     parser = argparse.ArgumentParser()
@@ -99,7 +97,6 @@ def main():
         )
         
         _, pendf_MTs, _ = tp.extract_endf_specs(pendf_path)
-        pendf_MTs = array(pendf_MTs)
         gas_MTs = set(pendf_MTs) & set(range(GAS_MT_MIN, GAS_MT_MAX))
         MTs |= {int(gas_MT) for gas_MT in gas_MTs}
 


eitan-weinstein:
I was able to process through the whole of TENDL2017 through this new version of `ALARAJOYWrapper` and follow through with the data library conversion and single initial nuclide test scripts. Just looking at gas production only, initial results are showing that we are successfully tracking gas production to some level, but tritium production is still down by roughly an order of magnitude from FENDL2 (as well as <sup>3</sup>He). 

image

I'll start looking tomorrow morning to see if I can figure out why this is the case.

eitan-weinstein:
Per conversations during the fusion activation meeting yesterday, this tritium underrepresentation may actually be an appropriate adjustment, based on updated (n,Xt) cross-sections from FENDL2 to newer TENDL updates (borrowing Tim's plots from our meeting 10/9/25):

plotTritProdFe56fendl21vsTENDL2015

plotspecificActivityRatio

The second plot (which is using FENDL2 data for ALARA) is actually overrepresenting tritium's contributions to specific activity at 100 years by >50%, so actually this new tritium count with ALARAJOY brings ALARA much closer to FISPACT's data at the century scale.

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -295,20 +302,27 @@ def run_njoy(element, A, matb):
...
+        (dir / fileinfo['dir']).mkdir(exist_ok=True)
 
-    return file_metadata['GENDF']['save'], result.stderr
+        fileinfo['save'] = save_path.with_suffix(fileinfo['ext'])
+        tape_save = Path(f'tape{fileinfo['tape']}')
+        if fileinfo['ext'] == '.gendf':
+            tape_save.rename(fileinfo['save'])
+            ensure_gendf_markers(fileinfo['save'], matb)
+        else:
+            shutil.copy(tape_save, fileinfo['save'])
```
gonuke: I'm pretty sure we can use `Path` for this

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -140,12 +139,18 @@ def fill_input_template(material_id, MTs, element, A, mt_dict, temperature):
...
+            group-structured GENDF).
+            (Defaults to None)
     
     Returns:
         template (str): Modified template with the material-
             specific information substituted in for the $identifiers,
             converted to a string.
     """
+    
+    inp = groupr_input if run_type == 'GROUPR' else njoy_prep_input
```
gonuke: Why not just pass in the template itself?

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -295,20 +302,27 @@ def run_njoy(element, A, matb):
...
-            save_path = dir / fileinfo['dir'] / f'tendl_2017_{element}{str(A).zfill(3)}'
-
-            # Ensure existence of save directory for PENDF/GENDF files
-            (dir / fileinfo['dir']).mkdir(exist_ok=True)
+        fileinfo = file_metadata[file_capture]
+        save_path = (
+            dir / 
+            fileinfo['dir'] / 
+            f'tendl_2017_{element}{str(A).zfill(3)}'
+        )
```
gonuke:
```suggestion
        fileinfo = file_metadata[file_capture]
        save_path = dir / fileinfo['dir']
        save_path.mkdir(exist_ok=True)
        save_path = save_path / f'tendl_2017_{element}{str(A).zfill(3)}'
```

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -295,20 +302,27 @@ def run_njoy(element, A, matb):
...
 
-            fileinfo['save'] = save_path.with_suffix(fileinfo['ext'])
-            Path(f'tape{fileinfo['tape']}').rename(fileinfo['save'])
-            if fileinfo['ext'] == '.gendf':
-                ensure_gendf_markers(fileinfo['save'], matb)
+        # Ensure existence of save directory for PENDF/GENDF files
+        (dir / fileinfo['dir']).mkdir(exist_ok=True)
```
gonuke: Suggested this be moved above...

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -324,7 +338,7 @@ def cleanup_njoy_files(output_path = dir / 'njoy_ouput'):
         None
     """
 
-    intermediate_files = [INPUT] + [Path(f'tape{i}') for i in range(20,25)]
+    intermediate_files = [INPUT] + [Path(f'tape{i}') for i in range(20,26)]
```
gonuke: Magic numbers: can we define 20 & 26 (or 25, as the case may be)

gonuke (changes requested):
A few suggestions and queries

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -295,20 +304,23 @@ def run_njoy(element, A, matb):
...
+        fileinfo['save'] = save_path.with_suffix(fileinfo['ext'])
+        tape_save = Path(f'tape{fileinfo['tape']}')
+#        tape_save.rename(fileinfo['save'])
+        if fileinfo['ext'] == '.gendf':
+            tape_save.rename(fileinfo['save'])
+            ensure_gendf_markers(fileinfo['save'], matb)
+        else:
+            shutil.copy(tape_save, fileinfo['save'])
+            
+        return fileinfo['save'], result.stderr
```
gonuke: You return `result.stderr` but this branch only runs  `if not result.stderr`?

tools/ALARAJOYWrapper/njoy_tools.py
gonuke: Seems like something we should be able to figure out.

It looks like when there does happen to be an error, this method doesn't actually return anything.  ie. there is no `return` line when an error occurs and therefore, we don't get much insight about what the error is.

gonuke (changes requested):
I'd like to figure out what's going on when the NJOY run fails.  Perhaps the return line needs to be indented one less?

PR #138, opened 2025-10-27
Expanding NJOY Usage to Create PENDF Files as part of NJOY Run (Clean)

eitan-weinstein:
Supplants #125  (rebase issues).

> Closes https://github.com/svalinn/ALARA/issues/118
>Closes https://github.com/svalinn/ALARA/issues/119

>This PR modifies the ALARAJOYWrapper to use NJOY more robustly by creating its PENDF files in house, instead of relying on their availability from the TENDL2017 data base. There are a number of key isotopes in TENDL 2017 that only have a TENDL file available, rather than a preprocessed PENDF as well. In order to convert to the groupwise cross-section data needed for ALARA, NJOY's GROUPR module uses both a TENDL/ENDF and a PENDF file pair to perform this conversion. However, TENDL 2017 provides the NJOY input files uses to produce said pointwise data, so by following the general format found for isotopes that do have available PENDFs, I expanded the njoy_input Template to accommodate for that.

>As an initial test, I have run this on 1H and 2H to confirm that isotopes that otherwise only have a TENDL file in the database can be converted and produce reasonable (although yet unvalidated) results. Moreover, I have tested this on 56Fe to start investigating and begin the verification process needed for https://github.com/svalinn/ALARA/issues/120 .

### DIFF
--- a/tools/ALARAJOYWrapper/njoy_tools.py
+++ b/tools/ALARAJOYWrapper/njoy_tools.py
@@ -12,15 +12,11 @@ def set_directory():
     Arguments:
         None
     Returns:
-        dir (pathlib._local.pathlib._local.PosixPath): Path to the current
-            working directory (CWD) from which the command was called.
+        dir (pathlib._local.PosixPath): Path to the current working directory
+            (CWD) from which the command was called.
     '''
 
-    dir = Path.cwd()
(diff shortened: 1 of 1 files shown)


gonuke:
I think this needs a rebase (or something) to capture #126

gonuke:
Another conflict in `groupr_tools.py`?

PR #140, opened 2025-10-31
Add input and results of activating all nuclides for 20 years

### LINKED ISSUE #104: Define a nominal ALARA scenario as the basis for comparison
Define a base problem that will be used to populate the database of results.

* every nuclide separately
* a single flux spectrum
* a single irradiation time

(This may look like #77)

This will be complete with an input file and supporting files added to the repo

anu1217:
fixes #104 

- This input file uses the FNSF FW spectrum used for the other two examples

### DIFF
--- a/examples/20y_all_nuc/20y_all_nuc.inp
+++ b/examples/20y_all_nuc/20y_all_nuc.inp
@@ -1,0 +1,1747 @@
+geometry rectangular
+
+volume
+     1    h:1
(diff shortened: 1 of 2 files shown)


gonuke (approved):
This looks good to me.  I assume (hope?) that you wrote a script to generate this input file.  I wonder if we should PR that script, too? (in a different PR)

PR #142, opened 2025-11-04
Comparison Tool Part 1: Simulation Tools for Single Element Comparison Tool

eitan-weinstein:
Responding to the need to break up #135 into smaller PRs, this starts by creating a new Python module `alarajoy_QA.py` in `ALARAJOYWrapper/` to be called within a Jupyter notebook `alarajoy_QA_notebook.ipynb`. Overall, following the work that I had started on in #135 , this notebook will allow the user to select any single element, run a simple ALARA irradiation scenario modelled after `examples/Results_singleElement_Fe_FENDL2/singleElement.ala` with multiple data sets (although with specific focus on comparing ALARAJOY-processed FENDL3.2x data against FENDL2 data). 

This PR includes functions needed to prepare and execute these ALARA runs, which is done iteratively in the Jupyter notebook. As currently structured, to be able to produce two ALARA output files to compare from each dataset, the user first has to follow the instructions in `./README.md` to use ALARAJOYWrapper and ALARA's `convert_lib` to produce an ALARA binary library from FENDL3.2x data. From there, and with the path to said converted library, the user needs only to input which element they are interested in irradiating.

### DIFF
--- a/tools/ALARAJOYWrapper/alarajoy_QA_notebook.ipynb
+++ b/tools/ALARAJOYWrapper/alarajoy_QA_notebook.ipynb
@@ -1,0 +1,85 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "1112b209",
+   "metadata": {},
+   "source": [
+    "# **ALARAJOY Library Conversion QA Notebook**\n",
+    "\n",
+    "\n",
+    "This Jupyter Notebook is designed to enable comparisons for neutron activation responses of a given single parent element as calculated by ALARA for the purpose of validating ALARAJOY-processed data from the FENDL3.2x data sets."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3298730d",
+   "metadata": {},
+   "source": [
+    "**Import Packages**"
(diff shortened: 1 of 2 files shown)


tools/ALARAJOYWrapper/alarajoy_QA_notebook.ipynb
```diff
@@ -0,0 +1,85 @@
...
+    "    'fendl3' : '../../examples/data/fendl3'\n",
+    "}\n",
+    "\n",
+    "element = input('Select single parent element to evaluate: ').lower()\n",
+    "\n",
+    "for libname, binary in data_soure.items():\n",
+    "    alara_input = qa.fill_alara_template(element, binary)\n",
+    "    qa.write_alara_input_file(alara_input)\n",
+    "    output = qa.run_alara(element,libname)\n",
+    "    data_soure[libname] = output"
```
gonuke: Do you intend to overwrite the `data_source` here?  Might make sense to store this in another dictionary.

gonuke (changes requested):
Only one little comment here...

PR #143, opened 2025-11-04
Comparison Tool Part 2: Loading in Tabular Data into Nested Dictionary.

eitan-weinstein:
Follow up to #142.

Following running of single element ALARA simulations with both FENDL2 and ALARAJOY-processed FENDL3.2x ALARA formatted binary libraries, this PR includes functions that utilize `../tools/alara_output_parser.py` to read in the tabular data from the ALARA output files. This data is parsed into Pandas DataFrames, which are stored within a dictionary of dictionaries containing all of the tables for both simulations, with each table containing its metadata of:

- Data source: e.g. 'fendl2', 'fendl3'
- Variable: e.g. 'Number Density', 'Specific Activity', etc.
- Unit: e.g. 'atoms/kg', 'Bq/kg', etc.
- Data: parsed tabular data in a Pandas DataFrame.


tools/ALARAJOYWrapper/alarajoy_QA_notebook.ipynb
```diff
@@ -0,0 +1,123 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "7bd99054",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "dfs = qa.process_metadata(data_soure)\n",
+    "\n",
+    "times = list(dfs[list(dfs.keys())[0]]['Data'].columns[1:])\n",
```
gonuke: Wonder if there is a more readable way to do this?

PR #144, opened 2025-11-04
Comparison Tool Part 3: Versatile Plotting Tool and Some Basic Implementations

eitan-weinstein:
Follow up to #143 .

This PR introduces the `plotting_single_response()` function to `alarajoy_QA.py`, which can produce plots of a given ALARA output variable against cooling times, either for a single data source or comparatively between multiple. The function includes a robust set of optional parameters to choose to plot specific daughter elements, filter out small contributors, show only total values, etc. 

For the actual example implementation of this function in the Jupyter notebook, however, I've included:

- A plot of a single data source's number density against time
- A plot comparing both data sources' number densities against time
- A plot comparing both data sources' gas production number densities against time
- Plots comparing total values for number density, specific activity, decay heat, and contact dose over time

### DIFF
--- a/tools/ALARAJOYWrapper/alarajoy_QA.py
+++ b/tools/ALARAJOYWrapper/alarajoy_QA.py
@@ -481,7 +481,7 @@ def construct_legend(ax, data_comp=False):
 
 #---------------------------- Plotting functions -----------------------------
 
-def plot_single_element(
+def plot_single_response(
         df_dicts, 
         total=False, 
         element='', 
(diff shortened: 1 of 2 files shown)


gonuke:
Rebase to capture #143?

tools/ALARAJOYWrapper/alarajoy_QA.py
```diff
@@ -207,4 +208,428 @@ def process_data(
...
+    for column in alara_df.columns[1:]:
+        if column == 'shutdown':
+            times.append(time_dict['shutdown'])
+        else:
+            time = column.split('y')[0]
+            times.append(float(time) * time_dict['y'])
+
+    return times
+
+def separate_total(alara_df):
```
gonuke: I wonder if we should update `alara_output_parser` to always ignore the total?  We can always recreate it by summing over a dataframe, and then don't have to worry about separating it...

PR #145, opened 2025-11-04
Comparison Tool Part 4: Pie Charts

eitan-weinstein:
Follow up to #144 .

Introduces the `single_data_source_pie()` function which allows for the creation of a pie chart for a particular data library's variable at a given cooling time. By iterating through the two data libraries, as well as each variable and time, the implementation of this function in the Jupyter notebook allows the user to get a detailed view at each time step to compare contributions of different nuclides, and benefit from insight to the way that isotopes that may not contribute greatly to number density at any given time can have highly significant impacts on other variables (with potentially insightful differences across the data sources).


gonuke:
The notebook doesn't load in the browser... does it work for you locally?

PR #146, opened 2025-11-05
Comparison Tool Part 5: Relative Plots

eitan-weinstein:
Follow up to #145.

This PR introduces a function `relative_contributions()` to `alarajoy_QA.py` that converts each cell to it's relative contribution to the total at each time. In the Jupyter notebook, this function is used prior to plotting again with `plot_single_response()`, with additional parameters specified for it being a relative plot, to expand the individual time slices represented in the pie charts to a cohesive view of a single variable's nuclide responses over time.


tools/alara_output_processing.py
```diff
@@ -65,13 +65,13 @@ def _parse_table_data(
...
+            df = pd.read_csv(
                 StringIO('\n'.join(current_table_lines)), sep=r'\s+'
-            ))
+            )
 
-            adf.columns = [c.replace('_', '') for c in adf.columns]
+            df.columns = [c.replace('_', '') for c in df.columns]
             key = f'{current_parameter} - {current_block}'
-            self.results[key] = adf
+            self.results[key] = ALARADFrame(df)
```
eitan-weinstein: When I was creating the `relative_contributions()` function, it was running into issues with the column operations from this method, but worked fine if I executed line 72 on a DataFrame, rather than an ALARADFrame. I'm not sure exactly *why* this is the case, but as far as the output from `_parse_table_data() is concerned, converting the DataFrame to an ALARADFrame after its initialization is all the same, and it appeared to be necessary for  proper column operations.

PR #148, opened 2025-11-06
Add option to convert cooling times in output file to seconds

### LINKED ISSUE #137: add option to print time in seconds in output tables
For readability, time is currently output in convenient units, e.g. day, year, etc.  It will be more machine readable if they are all output in seconds, which could be an option to users.

This could either be a command-line option, or an option in the `OutputFormat` block of the input file like the other units.

anu1217:
This PR introduces changes that allow the user to select the units (default or seconds) in which cooling times are printed in the headers of the output tables, if the `units` keyword is added in the `output` section of the input file.

Adding this in as a draft for now as this change has had many moving pieces, with changes to function definitions in multiple files, and I anticipate more work being necessary for everything to work as intended.

fixes #137


src/Loading.C
```diff
@@ -318,7 +320,7 @@ void Loading::tally(Result *volOutputList, double vol)
     what type of normalization is being used, so that the correct
     output information can be given. */
 void Loading::write(int response, int writeComp, CoolingTime* coolList, 
-		    int targetKza, int normType)
+		    int targetKza, int normType, const OutputFormat* outFmt)
```
gonuke: Why do you need `outFmt` here?  it doesn't appear to be used

And if you don't need it, then you don't need to include it's header above.

src/CoolingTime.C
```diff
@@ -131,61 +132,92 @@ int CoolingTime::makeCoolingTimes(double *& coolingTimes)
...
     }
-  cout << endl;
-  writeSeparator();
+
+    cout << endl;
+    writeSeparator();
 }
 
-void CoolingTime::getCoolTimesStrings(std::vector<std::string>& coolTimesList)
+void CoolingTime::getCoolTimesStrings(std::vector<std::string>& coolTimesList, const OutputFormat* outFmt)
```
gonuke: The header defines this as an integer not a a pointer to type `OutputFormat`.  I think I like that choice better since it means you don't need to test if `outFmt` is defined below.

src/OutputFormat.C
gonuke: I think we need to write the cooling time units for these ones, too, probably

src/Volume.C
```diff
@@ -617,7 +619,7 @@ void Volume::postProc()
     normalization is being used, so that the correct output information
     can be given. */
 void Volume::write(int response, int writeComp, CoolingTime* coolList, 
-		   int targetKza, int normType)
+		   int targetKza, int normType, const OutputFormat* outFmt)
```
gonuke: Why do you need `outFmt` here?  it doesn't appear to be used.

And if you don't need it, then you don't need to include it's header above.

src/Result.C
```diff
@@ -607,6 +614,16 @@ void Result::setNorm(double passedActMult, int normType)
   default:
     metricMult = 1;
   }
+  switch (cooltimeType) {
+  case COOLTIME_S:
+    cooltime_units = 2;
+    break;
+  case COOLTIME_DEF:
+    cooltime_units = 1;
+    break;
```
gonuke: Don't need these lines since the same thing happens in both cases

```suggestion
```

src/Result.C
```diff
@@ -607,6 +614,16 @@ void Result::setNorm(double passedActMult, int normType)
...
+  default:
+    cooltime_units = 1;
+  }
```
gonuke: Why not just:

```suggestion
  cooltime_units = cooltimeType;
```

gonuke (changes requested):
I'm a little surprised that this compiles, given some mismatches in the header and code.

anu1217:
Before marking this as ready for review, I changed my approach to this problem and pushed all my changes to a different branch, before merging the new branch into this one. I believe this didn't execute the way I expected and left some extraneous changes, which I think have all been resolved now. The folded dose/WDR still need to be modified to print out the cooling time in the desired units.

src/OutputFormat.C
```diff
@@ -175,7 +183,25 @@ OutputFormat* OutputFormat::getOutFmts(istream& input)
 	      next->normType = OUTNORM_CM3;
 	      break;
 	    }
-	  break;
+	  
+	  delete[] next->cooltimeUnits;
+	  input >> token;
+	  next->cooltimeUnits = new char[strlen(token)+2];
```
gonuke: For better or worse, this is not backwards compatible and will make old input files fail. We should discuss how important that is...

gonuke (changes requested):
Sorry - this review was 1/2 done a while ago and got lost...

anu1217 (commented):
I've addressed most of the feedback here and have tested these changes in a container.

src/OutputFormat.C
```diff
@@ -141,6 +149,7 @@ OutputFormat* OutputFormat::getOutFmts(istream& input)
       switch (1<<type)
 	{
 	case OUTFMT_UNITS:
+	  {
```
gonuke: Do we need these braces?

src/OutputFormat.C
```diff
@@ -175,7 +184,34 @@ OutputFormat* OutputFormat::getOutFmts(istream& input)
 	      next->normType = OUTNORM_CM3;
 	      break;
 	    }
-	  break;
+	  
+	  delete[] next->cooltimeUnits;
+	  next->cooltimeUnits = new char[strlen(token)+1];
```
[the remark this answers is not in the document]
gonuke: Perhaps we figure out a way to make it optional?

src/OutputFormat.C
```diff
@@ -175,7 +184,34 @@ OutputFormat* OutputFormat::getOutFmts(istream& input)
...
+			}
+		else if (tolower(token[0]) == 'd')
+			{
+			input.seekg(pos);
+			}
+		else
+			{
+			error(230, "Unknown cooling time unit '%s'", token);		
+			}
+		}
```
gonuke: Why might need to implement clang-formatting next, but this style is preferred:

```suggestion
	  if (input >> token) {
		if (tolower(token[0]) == 's') {
			next->cooltimeType = COOLTIME_S;

			delete[] next->cooltimeUnits;
			next->cooltimeUnits = new char[strlen(token)+1];
			strcpy(next->cooltimeUnits, token);
		} else if (tolower(token[0]) == 'd') {
			input.seekg(pos);
		} else {
			error(230, "Unknown cooling time unit '%s'", token);		
		}
	}
```

src/Result.C
```diff
@@ -592,12 +593,19 @@ void Result::readDump()
...
+
+
+void Result::setNorm(double passedActMult, int normType, int cooltimeType)
 {
 
   actMult = passedActMult;
 
-  switch (normType) {
+  switch (normType) 
+  {
```
gonuke:
```suggestion
  switch (normType) {
```

gonuke (changes requested):
Sorry for the delay - I had some pending changes that I forgot to submit.

src/OutputFormat.C
```diff
@@ -175,7 +184,34 @@ OutputFormat* OutputFormat::getOutFmts(istream& input)
...
+			{
+			next->cooltimeType = COOLTIME_S;
+
+			delete[] next->cooltimeUnits;
+			next->cooltimeUnits = new char[strlen(token)+1];
+			strcpy(next->cooltimeUnits, token);
+			}
+		else if (tolower(token[0]) == 'd')
+			{
+			input.seekg(pos);
```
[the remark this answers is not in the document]
anu1217: I've changed the logic used here, but since `while (strcmp(token,"end"))` is reading the list of output types until `end` has been found, doing nothing in this case would consume the `end` token earlier than we want

src/OutputFormat.C
anu1217: We need these braces due to the initialization of `std::streampos pos` in the case, or we get a "crosses initialization" error during compilation

src/OutputFormat.C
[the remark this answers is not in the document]
anu1217: It should be optional now since any token (or lack thereof) that doesn't start with "s" is taken to be the default case

src/OutputFormat.C
```diff
@@ -175,7 +184,28 @@ OutputFormat* OutputFormat::getOutFmts(istream& input)
...
+		} else {
+			input.seekg(pos);
+
+			next->cooltimeType = COOLTIME_DEF;
+			next->cooltimeUnits = new char[4];
+			strcpy(next->cooltimeUnits, "def");
+		}
+	}	
+	break;	
+	  }
```
gonuke, line 207: This approach is fragile. For example, it fails if a user does NOT include the optional cooling time unit, and the next entry in the output format block is `specific_activity` - something that might be quite common.

gonuke: Given the way that ALARA reads input files, I don't think we can avoid breaking backwards compatibility with this.

The only way might be to add a new optional line in the output format block just for this purpose.

src/OutputFormat.C
gonuke: So if we end up getting rid of the `streampos` (which my gut tells me we probably will) then we can get rid of these, too.

src/OutputFormat.C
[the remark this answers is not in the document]
gonuke: Right - but only if we try to make this new addition optional on this same `units` line.

gonuke (changes requested):
Seems we need to discuss whether this can be optional in this way.

PR #153, opened 2025-11-07
Create ALARA Post-Processor

### LINKED ISSUE #150: Create a python module with many single post-processing operations
Once the data for a single point in space over multiple resulting nuclides and times has been read into a dataframe, support a variety of simple operations:
* process time values into different units
* calculate the totals across columns
* filter for specific nuclides or elements
* aggregate small values into an "other" category
* (and possibly others)

eitan-weinstein:
Closes #150. 

This PR changes the current Python script `tools/alara_output_parser.py` to being a more robust postprocessor, `tools/alara_output_processing.py`. This is done by joining some of the pure data processing methods from `tools/ALARAJOYWrapper/alarajoy_QA.py` that have been merged already and from #144. 

To combine these two scripts, I also rewrote `alara_output_processing.py` as object-oriented, per our discussion yesterday (although we hadn't come to a firm conclusion on whether or not it would be necessary). I created two classes `TableParser` and `DataProcessing` to further separate concerns between the actual reading of the ALARA output data and then the handling of said data from the former QA methods. 

Currently, the `__main__` method only encapsulates the parsing functionality, more or less as it did before, with my thinking that that is something that can be valuable as a standalone usage in converting the output tables to CSVs, whereas the actual data processing would be best used within other applications, such as a new QA Jupyter notebook that I am going to work on in upcoming PRs or in other postprocessing applications that the user wants to create.

I'll note that this is my first time converting a program to an object-oriented paradigm, so it's possible that I've missed some programmatic conventions that I am unaware of, but I think in principle, this change is worthwhile to create this combined module.

### DIFF
--- a/tools/alara_output_processing.py
+++ b/tools/alara_output_processing.py
@@ -131,8 +131,8 @@ class FileParser:
     # ---------- Output ----------
     def write_csv_files(self):
         '''
-        Write out all DataFrames extracted from parsed ALARA output tables to
-            their own CSV files.
+        Write out all ALARADFrames extracted from parsed ALARA output tables
+            to their own CSV files.
 
         Arguments:
             self (alara_output_processing.FileParser): FileParser object.
@@ -188,7 +188,7 @@ class ALARADFrame(pd.DataFrame):
     def extract_totals(self):
         '''
         Select the values from the "total" row of an ALARA output table
-            DataFrame and write them out to a list.
+            ALARADFrame and write them out to a list.
         
         Arguments:
             self (alara_output_processing.ALARADFrame): Specialized ALARA
@@ -205,7 +205,7 @@ class ALARADFrame(pd.DataFrame):
     
     def filter_elements(self, elements):
         '''
-        Create a new DataFrame containing only the data for nuclides of a
+        Create a new ALARADFrame containing only the data for nuclides of a
             selected element or elements.
         
         Arguments:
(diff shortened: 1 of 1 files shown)


tools/alara_output_processing.py
```diff
@@ -0,0 +1,399 @@
+import pandas as pd
+import re
+import argparse
+from io import StringIO
+
+class TableParser:
```
gonuke: This is probably more of a `FileParser`

tools/alara_output_processing.py
```diff
@@ -0,0 +1,399 @@
...
+        
+        dfs = {}
+
+        if isinstance(data_source, pd.DataFrame):
+            dfs.update(
+                cls.make_entry(
+                    inp_datalib, inp_variable, inp_unit, data_source
+                )
+            )
+            return dfs
```
gonuke: Maybe we just have users call `make_entry` directly in these cases, and only use this method for processing multiple output files?  Then we could have one method called `make_entry` and one called `make_entries`? Or something like that...

tools/alara_output_processing.py
```diff
@@ -0,0 +1,399 @@
...
+        
+        Returns:
+            None
+        '''
+
+        for key, df in self.results.items():
+            filename = self.sanitize_filename(key) + '.csv'
+            df.to_csv(filename, index=False)
+
+class DataProcessing:
```
gonuke: If we separate out the `ALARADFrame` to its own class, then maybe this one can be called `DataLibrary`?

gonuke (changes requested):
I really like this change and thanks for introducing the object oriented approach - it will make future use of this all much cleaner!!!

I've made a few design suggestions based on these changes. Happy to discuss more if you have concerns.

Issue #151, opened 2025-11-06
Create a module for standard plots that one might perform with ALARA data


gonuke:
Assume that a user had read data into a dataframe using our tools and possibly processed that dataframe using our tools, provide some typical plots that allow users to inspect and compare results

PR #154, opened 2025-11-10
Updated Comparison Tool Parts 1-3: Refactored for New ALARA Output Parser

### LINKED ISSUE #168: Method to plot a single result from an `ALARADFrame`
A simple method can call `matplotlib` functions to create a line plot of results from a single nuclide as a function of cooling time, with standard axis labels. The plots should probably be log-log plots.

### LINKED ISSUE #169: Method to plot multiple time series results on the same plot
As an extension of #168, a dataframe with multiple rows can be provided to plot multiple lines on a single axes.  The same standard axis labels and log-log formatting can be used.

We might consider whether it's possible to automate the process of grouping the different entries so that the multiple rows could refer to:
* only multiple nuclides
* only multiple cases
* only multiple geometry indices
* multiple nuclides over multiple cases
* multiple nuclides over multiple geometric locations
* multiple cases over multiple geometric locations

It is probably too complicated to imagine a plot with multiple nuclides, multiple cases, and multiple geometric locations.

eitan-weinstein:
Closes #168 .
Closes #169 .

This PR implements changes made to `tools/alara_output_parser.py` for `alarajoy_QA`, as far as the most up-to-date version, as well as the newest plotting suggestions otherwise in #144 . This PR partially completes #151, however, I will be making adjustments to PRs #145, #146, and #147 as well to get them up to date with the new methods developed in #153.

### DIFF
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -58,17 +58,25 @@ def preprocess_data(
     }
     if nuclides:
         filter_dict['nuclide'] = nuclides
-        
-    filtered = adf.filter_rows(filter_dict)
     
+    filtered = adf.filter_rows(filter_dict)
+
+    preset_time_unit = filtered['time_unit'].unique()[0]
+    if time_unit != preset_time_unit:
+        filtered['time'] = filtered['time'].transform(
+            lambda v: aop.convert_times(
+                array([v]), from_unit=preset_time_unit, to_unit=time_unit
+            )[0]
+        )
+        filtered['time_unit'] = [time_unit] * len(filtered)
+        
+    times = sorted(filtered['time'].unique().tolist())
+
     piv = filtered.pivot(
         index='nuclide',
         columns='time',
         values='value'
     )
-    times = sorted(aop.convert_times(
-        array(adf['time'].unique().tolist()), 's', time_unit
-    ))
 
     if sort_by_time:
         sort_by_time = aop.extract_time_vals([sort_by_time])[0]
--- a/tools/ALARAJOYWrapper/alarajoy_QA_notebook.ipynb
+++ b/tools/ALARAJOYWrapper/alarajoy_QA_notebook.ipynb
@@ -21,7 +21,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 12,
+   "execution_count": 78,
    "id": "57313772",
    "metadata": {},
    "outputs": [],
@@ -46,7 +46,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 13,
+   "execution_count": 79,
    "id": "4eae0d85",
    "metadata": {},
    "outputs": [
@@ -56,7 +56,7 @@
        "{'fendl2': 'fe_fendl2.out', 'fendl3': 'fe_fendl3.out'}"
       ]
      },
-     "execution_count": 13,
+     "execution_count": 79,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -103,7 +103,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 14,
+   "execution_count": 80,
    "id": "7bd99054",
    "metadata": {},
    "outputs": [
@@ -307,7 +307,7 @@
        "[2478 rows x 9 columns]"
       ]
      },
-     "execution_count": 14,
+     "execution_count": 80,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -330,7 +330,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 29,
+   "execution_count": 87,
    "id": "1e025489",
    "metadata": {},
    "outputs": [
@@ -370,7 +370,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 30,
+   "execution_count": 88,
    "id": "1c36abaa",
    "metadata": {},
    "outputs": [
@@ -398,7 +398,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 31,
+   "execution_count": 89,
    "id": "7396985f",
    "metadata": {},
    "outputs": [
@@ -445,7 +445,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 32,
+   "execution_count": 91,
    "id": "70ef1c57",
    "metadata": {},
    "outputs": [


gonuke:
Does this need a rebase for the `output_parser` additions?

tools/ALARAJOYWrapper/alarajoy_QA.py
```diff
@@ -112,99 +109,242 @@ def run_alara(element, libname):
...
+        isotope (str): Reformatted identifier of the isotope of the form
+            ᴬelement.
+    '''
+
+    if isotope == 'total' or isotope == 'Other':
+        return isotope
+    else:
+        element, A = isotope.split('-')
+        element = element.capitalize()
+        return f'$^{{{A}}}\\mathrm{{{element}}}$'
```
gonuke: This might be simpler and equivalent since `mathrm` is basically the same as non-math
```suggestion
        return f'$^{{{A}}}${element}'
```

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -0,0 +1,285 @@
...
+        label (str): Series label generated from 
+            ax.get_legend_handles_labels().
+    
+    Returns:
+        isotope (str): Isotope being plotted.
+        run_lbl (str): Run label of plotted isotope. Will be empty string
+            for single run plotting.
+    '''
+
+    if '(' in label:
```
gonuke: Can you add a comment about the assumed format of the label if this condition is true?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -0,0 +1,285 @@
...
+                times (list of floats): Cooling times (columns of adf).
+    '''
+
+    all_labels = set()
+    all_data = []
+    for df_dict in df_dicts:
+        adf = df_dict[data_key]
+        times = adf.process_time_vals(seconds=seconds)
+        adf = adf.T
+        for col in adf.columns:
```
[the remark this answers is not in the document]
gonuke: If we didn't transpose first, could we just perform this operation on all the entries in the first column?  The transpose is not a necessary step yet (although it may make life easier for plotting)

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -0,0 +1,285 @@
...
+            }
+        data_key (str, optional): Keyword for df_dicts to direct to the
+            desired ALARADFrame to source the plot's data.
+            (Defaults to 'Data')
+        seconds (bool, optional): Option to convert cooling times from
+            years to seconds.
+            (Defaults to True)
+
+    Returns:
+        all_data (list of tuples): A list of all relevant data for each table
```
gonuke: What if instead of returning a list of tuples, this method added a new entry for the `times` to the `df_dict` for each entry in `df_dicts`, and collected the labels to make the `color_map`.

Notwithstanding the discussion about putting some of the metadata into the tables, I like the idea of having the `df_dicts` cache the processed metadata rather than making new data structures that may be less transparent.

eitan-weinstein:
> It looks like you tried a rebase here, but all the changes from the previous PR are still here. Maybe its easier to just merge main into this branch.

I just rebased again to `upstream/main` and I think it looks right now with the most recent changes.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -0,0 +1,333 @@
...
+        columns='time',
+        values='value'
+    )
+    times = aop.convert_times(
+        array(adf['time'].unique().tolist()), 's', time_unit
+    )
+
+    if head:
+        sort_by_time = aop.extract_time_vals([sort_by_time])[0]
+        piv = piv.sort_values(sort_by_time, ascending=False).head(head)
```
gonuke: 1. why does this sorting only happen if `head`?
2. the sorting would be a good stand alone function as I anticipate a user wanting to sort a dataframe and although it's only a couple of lines, it's not super obvious how

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -0,0 +1,333 @@
...
+
+    Returns:
+        color_map (dict): Preset color map for each nuclide in the pivot table
+            indices.
+    '''
+
+    all_nucs = set()
+    for piv in list_of_pivs:
+        for nuc in piv.index:
+            all_nucs.add(nuc)
```
gonuke:
```suggestion
        all_nucs.add(set(piv.index))
```

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -0,0 +1,333 @@
...
+
+    nucs_sorted = sorted(all_nucs)
+    cmap = plt.cm.get_cmap('Dark2')
```
gonuke: Maybe add a cmap as an input to this function, with default of `'Dark2'`?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -0,0 +1,333 @@
...
+        isotope (str): Reformatted identifier of the isotope of the form
+            ᴬelement.
+    '''
+
+    if isotope == 'total' or isotope == 'Other':
+        return isotope
+    else:
+        element, A = isotope.split('-')
+        element = element.capitalize()
+        return f'$^{{{A}}}${element}'
```
gonuke: my hero! 🦸‍♂️

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -0,0 +1,333 @@
...
+            nuclides=nuclides,
+            time_unit=time_unit,
+            sort_by_time=sort_by_time,
+            head=head,
+        )
+        data_list.append((run_lbl, times, filtered, piv, linestyle))
+
+    color_map = build_color_map([piv for (_, _, _, piv, _) in data_list])
+    for run_lbl, times, filtered, piv, linestyle in data_list:
+
```
gonuke:
```suggestion

    for run_lbl, times, filtered, piv, linestyle in data_list:
```

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -0,0 +1,333 @@
...
+            run_lbl=run_lbl,
+            variable=variable,
+            nuclides=nuclides,
+            time_unit=time_unit,
+            sort_by_time=sort_by_time,
+            head=head,
+        )
+        data_list.append((run_lbl, times, filtered, piv, linestyle))
+
+    color_map = build_color_map([piv for (_, _, _, piv, _) in data_list])
```
gonuke:
```suggestion
    color_map = build_color_map([data[3] for data in data_list])
```

PR #161, opened 2025-11-17
Reorganizing Postprocessor to Store all Data in Single ADF

### LINKED ISSUE #158: Update postprocessor to put all data in a single dataframe
Rather than creating dictionaries of data frames with metadata encoded in the dictionary keys, put all the data in a single dataframe and add columns for metadata:
1. user-specified label
2. output variable type
3. geometric block type
4. geometric block number

### LINKED ISSUE #164: Update structure of `ALARADFrame` to a flat canonical structure
Data read from ALARA output files should be stored in a flat table with a single result per row, and columns:
* time
* nuclide
* geometry index
* geometry type
* variable index
* user label
* value

This will allow all data from multiple cases to be stored in the same data frame and extracted for additional operations as necessary.

### LINKED ISSUE #165: Create a method to filter an `ALARADFrame`
With output data in a canonical form in an `ALARADFrame` it should be straightforward to extract data using a filter.

A generic filter function can take multiple keys that correspond to columns in the `ALARADFrame` and each key can have a single value or list of values.  Panda routines can be used to perform the matching of the data in the appropriate keys.

The filter key/value pairs could be passed in as a single dictionary or as `kwargs` following other standard patterns.

Before each key is applied in the filter, it should be checked to ensure that it is a column in the table and ignored with a warning if not.

### LINKED ISSUE #166: Allow nuclides to be filtered by element
The basic filtering added to close #165 will allow filtering by individual nuclides. It may be desirable for users to be able to filter based on elements, so additional logic can be added to expand the filtering the support this.

### LINKED ISSUE #167: Methods to convert time units
ALARA output includes cooling times in formats that are convenient for human readers but less so for automated parsing.

A method should be added that can convert one or more time entries from arbitrary units to seconds. This method can be used to store values in an `ALARADFrame` in units of seconds.

Another method could be added to convert seconds to convenient units as requested by users.

eitan-weinstein:
Closes #158 .
Closes #164 .
Closes #165 .
Closes #166 .
Closes #167 .

This PR updates `alara_output_processor.py` to store all tabular data from each run, table, block, and variable into a single `ALARADFrame` indexed by cooling times. Updates were made to other functions in the `ALARADFrame` class as well to modify their usage to this new paradigm and to give users flexibility in how they want to organize their data after initial processing.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -68,13 +68,13 @@ The columns for `adfs` are:
 Below is the example `head()` of an `ALARADFrame`:
 
 
-| | time | nuclide | run_lbl | block | block_num | variable | value |
-|-|------|---------|---------|-------|-----------|----------|-------|
-| 0 | 0.000000e+00 | h-1 | fendl2 | 0 | 1 | 0 | 1.176100e+22
-| 1 | 3.153600e+02 | h-1 | fendl2 | 0 | 1 | 0 | 1.176100e+22
-| 2 | 3.153600e+05 | h-1 | fendl2 | 0 | 1 | 0 | 1.176100e+22
-| 3 | 3.153600e+07 | h-1 | fendl2 | 0 | 1 | 0 | 1.176100e+22
-| 4 | 3.153600e+09 | h-1 | fendl2 | 0 | 1 | 0 | 1.176100e+22
+| | time | time_unit | nuclide | run_lbl | block | block_num | variable | var_unit | value |
+|-|------|-----------|---------|---------|-------|-----------|----------|----------|-------|
+| 0 | 's' | 0.000000e+00 | h-1 | fendl2 | 0 | 1 | 0 | atoms/kg | 1.176100e+22
+| 1 | 's' | 3.153600e+02 | h-1 | fendl2 | 0 | 1 | 0 | atoms/kg | 1.176100e+22
+| 2 | 's' | 3.153600e+05 | h-1 | fendl2 | 0 | 1 | 0 | atoms/kg | 1.176100e+22
+| 3 | 's' | 3.153600e+07 | h-1 | fendl2 | 0 | 1 | 0 | atoms/kg | 1.176100e+22
+| 4 | 's' | 3.153600e+09 | h-1 | fendl2 | 0 | 1 | 0 | atoms/kg | 1.176100e+22
 
 The five rows in the head correspond to the number density of <sup>1</sup>H in the 1st interval of a run of FENDL2 data with four cooling times.
 
@@ -89,9 +89,9 @@ filtered_adf = adf.filter_rows(
     }
 )
 ```
-The parameter `filter_dict` allows filtering over any number of columns and any number of filters per column, so long as multi-filters are input as a list. Filters are case-sensitive. Similarly to the Pandas parameter, `inplace` allows the user to decide whether to operate on the original `ALARADFRame` itself, or create a new frame.
+The parameter `filter_dict` allows filtering over any number of columns and any number of filters per column, so long as multi-filters are input as a list. Filters are case-sensitive.
 
-**Note:** When filtering the `nuclide` column, `ALARADFrame.filter_rows()` has functionality to select all nuclides of a particular element, as well as selecting individual nuclides. To do so, instead of  `filter_dict["nuclide"] = "fe-55"`, write `filter_dict["nuclide"] = "fe"` to filter all iron isotopes, instead of just <sup>55</sup>Fe, for example. Similarly, multiple whole elements can be selected by inputting them as a list for `filter_dict["nuclide"]`, however, filtering by whole elements and individual nuclides cannot be done in the same filtering operation.
+**Note:** When filtering the `nuclide` column, `ALARADFrame.filter_rows()` has functionality to select all nuclides of a particular element, as well as selecting individual nuclides. To do so, instead of  `filter_dict["nuclide"] = "fe-55"`, write `filter_dict["nuclide"] = "fe"` to filter all iron isotopes, instead of just <sup>55</sup>Fe, for example. Similarly, multiple whole elements can be selected by inputting them as a list for `filter_dict["nuclide"]`. It is also possible to filter by a combination of whole elements and individual nuclides.
 
 Below is an example filtering operation on the same `adf` from the above example:
 ```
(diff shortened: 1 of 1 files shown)


tools/alara_output_processing/alara_output_processing.py
```diff
@@ -130,57 +168,118 @@ def extract_tables(self):
...
     def _constructor(self):
         return ALARADFrame
+    
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.is_pivot = False
+        self.is_single_var = False
+        self.is_single_run = False
+
+    def select_single_var(self, variable):
```
gonuke: Could we combine these methods into a generic filtering method that takes a dictionary of different filters, e.g:
```
def filter_rows(self, filter):
    # take the intersection of the filter keys with the allowable columns
    # perform a filtering of the dataframe with all the provided/allowable columns
    # return that dataframe without pivoting yet
```

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -61,42 +63,78 @@ def _parse_table_data(
...
+        current_table_rows = []
+        df = pd.read_csv(
+            StringIO('\n'.join(current_table_lines)), sep=r'\s+'
+        ).set_index('isotope')
+
+        df.columns = [c.replace('_', '') for c in df.columns]
+        block_name, block_num_trail = current_block.split(' #')
+        variable = current_parameter.split(' [')[0]
+        adf = ALARADFrame(df)
+        adf.columns = adf.process_time_vals(parsing=True)
```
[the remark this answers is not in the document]
eitan-weinstein: They do not -- the index is not treated as a column

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -130,193 +164,223 @@ def extract_tables(self):
...
-            }
-        }
-
-    @classmethod
-    def make_entries(cls, runs_dict):
+        try:
+            totals = self.filter_rows({'nuclide : total'})['value'].tolist()
+        
+        except AttributeError:
+            totals = self[self.index == 'total'].iloc[0].tolist()
```
gonuke: I'm not sure I understand the difference between these?

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -130,193 +164,223 @@ def extract_tables(self):
...
+
+        else:
+           warn('ALARADFrame missing "total" row.')
+           totals = []
```
gonuke: Can't we calculate totals and add them to the ADF? (and return them)

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -61,42 +63,74 @@ def _parse_table_data(
...
+                    'value' : Float value for the corresponding variable
+                } 
+        '''
+
+        current_table_rows = []
+        df = pd.read_csv(
+            StringIO('\n'.join(current_table_lines)), sep=r'\s+'
+        ).set_index('isotope')
+
+        df.columns = [c.replace('_', '') for c in df.columns]
```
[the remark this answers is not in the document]
eitan-weinstein: After being read in, the column headers look like this:

```
Index(['shutdown', '1e-05_y', '0.01_y', '1_y', '100_y', '10000_y'], **dtype='object')
```

Looking at this in context of your comment on `extract_time_vals()`, it might make more sense to just update that method to just `split` at the underscore, and then whatever is after the underscore gives the units and makes for greater flexibility anyways.

tools/alara_output_processing/alara_output_processing.py
gonuke: Can we interrogate the way a table is oriented without a try-except?

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -3,12 +3,121 @@
...
+            numeric_times.append(0.0)
+        else:
+            time_str, unit = time.split('_')
+            numeric_times.append(float(time_str))
+    
+    return convert_times(
+        array(numeric_times), from_unit=unit, to_unit=to_unit
+    )
+
+def aggregate_small_percentages(piv, relative=False, threshold=0.05):
```
gonuke: I've proposed in the issues that we might want to separate the process of making a table into relative quantities rather than doing that on the fly in methods like this.

Let me know what you think about pulling out that feature from this method.

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -3,12 +3,121 @@
...
+        times (numpy.ndarray): 1-D NumPy array of the converted ALARA cooling
+            times.
+    '''
+    
+    numeric_times = []
+    for time in times:
+        if 'shutdown' in time:
+            numeric_times.append(0.0)
+        else:
+            time_str, unit = time.split('_')
```
gonuke: What if you didn't put the underscore into this during `_normalize_header`?  It seems that you insert the underscore there, and then promptly use it to split this value when you could just use the space?

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -61,42 +170,83 @@ def _parse_table_data(
...
+                    'block_num' : Geometric position of block,
+                    'variable' : ALARADFrame variable integer enumerator,
+                    'unit' : Unit for corresponding variable,
+                    'value' : Float value for the corresponding variable
+                } 
+        '''
+
+        # Prepare table text for csv.DictReader
+        text = '\n'.join(current_table_lines)
+        comma_delim = re.sub(r'\t+|\s{2,}', ',', text.strip())
```
gonuke: Is this maybe why you put the underscore in the times in the header, so that they don't match this regex?

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -130,193 +281,155 @@ def extract_tables(self):
...
-        other_row[self.columns[0]] = 'Other'
-        other_row[cols] = self.loc[small_mask, cols].sum()
-
-        mask_adf = self.loc[~small_mask].reset_index(drop=True)
-        if other_row[cols].sum() > 0:
-            mask_adf = pd.concat(
-                [mask_adf, pd.DataFrame([other_row])], ignore_index=True
-            )
+        # Case 1: Pivoted ALARADFrame with "total" row
+        if self.index.name == 'nuclide' and 'total' in self.index:
```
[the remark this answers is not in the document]
gonuke: We can then consider if we need a separate function to calculate totals in extracted/pivoted dataframes that are not member functions of ADF.  See #175 for discussion of this.

tools/alara_output_processing/alara_output_processing.py
eitan-weinstein: I'm starting to think that it might be valuable to just automatically calculate these relative values and include it as a column in the canonical structure so that any `ALARADFrame` produced will have this data (and the user can ignore the column if they're not interested in it). Thoughts?

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -130,193 +290,161 @@ def extract_tables(self):
...
+                raise ValueError(
+                    'Relative values can only be calculated for single run, '\
+                    f'block, and variable ALARADFrames. \nFilter {col} to '\
+                    'single value before calculating relative values.'
+                )
 
-        return mask_adf
+        totals = self.filter_rows({'nuclide' : 'total'})['value'].to_numpy()
+        times = sorted(set(self['time']))
+        total_map = dict(zip(times, totals))
```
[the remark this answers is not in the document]
gonuke: I think this will give you a numpy array of times and and values:

```suggestion
        totals = self.filter_rows({'nuclide' : 'total'})[['time','value']].to_numpy()
        total_map = { time: value for time, value in totals }
```

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -130,193 +290,161 @@ def extract_tables(self):
...
+        adf_rel = self[self['nuclide'] != 'total'].copy()
 
+        adf_rel['rel_value'] = adf_rel.apply(
```
gonuke: Since you've made a copy, I think it's OK to just change the `value` and return this as a separate DF with only the relative values.

gonuke (changes requested):
Further clarifying the intended interface for some of these functions.

In particular, I think it's reasonable for `calculate_relative_vals` to return a new DataFrame in which the `value` column is just the relative values, rather than adding a column.

If so, the `aggregate_small_percentages` can be expanded to aggregate small values with a threshold that could be absolute for  a dataframe that has not be converted to relative and could be relative for a dataframe that has.

tools/README.md
```diff
@@ -25,45 +26,130 @@ alara_output_processing -f /path/to/alara/output.txt
...
+    - "Contact Dose"
+    
+    The respective enumerator values can be accessed through `ALARADFrame().VARIABLE_ENUM[variable]`, where `variable` is one of the above variables.
+* `var_unit`: Unit for the corresponding variable.
+* `value`: Numeric value of the particular variable respone.
+
+Below is the example `head()` of an `ALARADFrame`:
+
+
+| | time | nuclide | run_lbl | block | block_num | variable | value |
```
gonuke: I think you need to add a `time_unit` and `var_unit` column to this table

tools/README.md
```diff
@@ -25,45 +26,130 @@ alara_output_processing -f /path/to/alara/output.txt
...
-}
+filtered_adf = adf.filter_rows(
+    filter_dict={
+        Column 1 : Value or [Value1, ..., Value N],
+        ...
+        Column N : Value or [Value1, ..., Value N]
+    }
+)
+```
+The parameter `filter_dict` allows filtering over any number of columns and any number of filters per column, so long as multi-filters are input as a list. Filters are case-sensitive. Similarly to the Pandas parameter, `inplace` allows the user to decide whether to operate on the original `ALARADFRame` itself, or create a new frame.
```
gonuke: I think we got rid of the `inplace` usage

gonuke (approved):
Thanks for sticking through this @eitan-weinstein

PR #162, opened 2025-11-20
Add ALARA cases with database-relevant scenarios

anu1217:
This PR adds in input and output files that are intended to become part of a database that explores different combinations of a given irradiation history. These files draw on the example of an ITER FW, for which beryllium was chosen in the past and tungsten is currently the primary candidate material. These examples are based on a pulse and dwell time of 400 s and 1400 s, respectively.

fixes svalinn/activationDB#1


anu1217:
Closing as these cases will be modified and PR'd to a separate repository.

Issue #163, opened 2025-11-20
Print details of individual pulse histories to output


anu1217:
Currently, the output prints the times and units of the entries in the schedule block, along with the delay time between the entries. The details of the associated pulse history (i.e. the number of pulses and delay between individual pulses) are not visible, which makes it so that total irradiation time cannot be inferred from the output alone.

anu1217:
> It is useful to think about how to write this information in a way that is both useful to human readers and doesn't require to much complex parsing for computer readers.

Agreed -- I can see this information being conveyed in a way that's descriptive and verbose (e.g. "pulsehistory [pulse name] with x pulses and delay y"), added to the schedule blocks that the pulsehistory appears in. Or it could be a table (or multiple?) with the schedule and pulse information. Do you have a preference/other ideas?

PR #177, opened 2025-11-24
Add new sample problems with increasing t_irr complexity

### LINKED ISSUE #176: Add example problem with varying pulse histories
So far, we only have sample problems with single steady state pulses.

It would be useful to add two or three problems with growing complexity of histories:

1. [ ] simple set of regular pulses
2. [ ] multi-level pulse history with fixed pulse width
3. [ ] multiple different pulses arranged in varying schedules

anu1217:
These three new sample problems are intended to set up irradiation histories that grow in complexity, in order to serve as a basis problem for addressing #163.

- sample10 contains 3 schedule blocks that each use a separate pulse history. All schedule and pulse histories are single level.
- sample11 contains 3 schedule blocks that all use 1 single pulse history. This pulse history contains 3 levels of pulsing.
- sample12 contains 4 schedule blocks and 4 pulse histories. The 4th pulse history is multi-level and the rest are single-level. The 3rd schedule block is multi-level and incorporates one of the single-level schedules as a sub-schedule. The 4th schedule block is multi-level and incorporates all 4 pulse histories.

fixes #176

### DIFF
--- a/sample/sample10
+++ b/sample/sample10
@@ -26,8 +26,8 @@ end
 flux  my_flux       ../examples/ref_flux_files/fluxfnsfIBfw_518MW.txt 1     0      default
 
 schedule top_sched
-    sch_1   once   0 s
-    sch_2   once   0 s
+    sch_1   once   1 d
+    sch_2   once   25 d
     sch_3   once   0 s
 end
 
--- a/sample/sample11
+++ b/sample/sample11
@@ -26,8 +26,8 @@ end
 flux  my_flux       ../examples/ref_flux_files/fluxfnsfIBfw_518MW.txt 1     0      default
 
 schedule top_sched
-    sch_1   once   0 s
-    sch_2   once   0 s
+    sch_1   once   1 h
+    sch_2   once   90 d 
     sch_3   once   0 s
 end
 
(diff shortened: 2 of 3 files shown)


sample/sample10
```diff
@@ -0,0 +1,56 @@
...
+data_library alaralib fendl2bin
+
+output zone
+        number_density
+end
+
+#     flux name    fluxin file   norm   shift   unused
+flux  my_flux       ../examples/ref_flux_files/fluxfnsfIBfw_518MW.txt 1     0      default
+
+schedule    sch_1
```
gonuke: I don't think these work the way you expect.  If you want all of these schedules to occur in sequence, I think they all need to be combined in a top level schedule
```
schedule top_sched
    sch_1   once   0 s
    sch_2   once   0 s
    sch_3   once   0 s
end

pulsehistory once
    1    0.0   s
end
```

anu1217: Experimented with different combinations of these schedules, and it looks like it is not just the top one that is being executed. Adding a new schedule and its corresponding pulse history changes the number densities of the nuclides. Somehow, the only entry that has no effect on the results (even when it is absent) is the second schedule/pulse history.

PR #178, opened 2025-11-25
Add a README file for the sample problems

gonuke:
In light of the possibility of adding more sample problems (#177), a REAMDE file is added to describe the basic features of each of the existing sample problems.

### DIFF
--- a/sample/readme.md
+++ b/sample/readme.md
@@ -155,7 +155,7 @@ volume 0.5 cm3 and the adjoint response defined in `adjfile2.gam` using 21 group
 ## sample9
 
 This example is used to demonstrate the (undocumented) `integrate_energy` option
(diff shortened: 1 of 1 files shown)


sample/readme.md
```diff
@@ -0,0 +1,176 @@
...
+
+The flux is read from the `fluxin1`, and normalized differently for each interval.
+
+The there is a single 10 year pulse.
+
+Results are produced at 11 cooling times the total heat
+and photon source, with units of Bq per cm3.
+
+Most initial nuclides will have their chains truncated with a tolerance of 1e-6,
+but impurities (defined as nuclides that are less than 10 ppm) will be truncated
```
anu1217:
```suggestion
but impurities (defined as nuclides with an initial concentration of less than 10 ppm) will be truncated
```

PR #179, opened 2025-12-02
Comparison Tool 4.5: Grid of Pie Charts across Cooling Times

### LINKED ISSUE #174: Grid of multiple pie charts
There may be times when it is useful to generate a grid of pie charts, where each pie chart represents a single time, variable, geometric index, and case, and the different rows and columns of the grid represent various times, variables, geometric indices, or cases.

This method could call the single pie chart method (#173) or perhaps not if it is better to decorate them collectively than individually.

eitan-weinstein:
Closes #174.

Expanding on pie chart functionality added in #145 to create a grid of pie charts for all cooling times for a given run and response variable. An example of what one of these grids looks like is attached:

f2566abd-cc67-4bfb-8073-c92917a88e47

### DIFF
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -592,7 +592,11 @@ def multi_time_pie_grid(
     threshold=0.05,
     time_unit='s',
     ncols=2,
-    cmap_name='Dark2'
+    cmap_name='Dark2',
+    individual_pie_dimension = 5.0, # equal width/height
+    horizontal_buffer = 3.0,
+    vertical_buffer = 0.7
+
 ):
     '''
     Create a grid of pie charts for a single ALARA response variable from a
@@ -623,6 +627,15 @@ def multi_time_pie_grid(
             the plots. Reference guide for Matplotlib Colormaps can be found
             at matplotlib.org/stable/gallery/color/colormap_reference.html
             (Defaults to "Dark2")
+        individual_pie_dimension (float, optional): Optional parameter to set
+            the width/height of each pie chart in grid.
+            (Defaults to 5.0)
+        horizontal_buffer (float, optional): Optional parameter to set a
+            buffer between each pie chart in a row.
+            (Defaults to 3.0)
+        vertical_buffer (float, optional): Optional parameter to set a buffer
+            between each pie chart in a column (as well as the legend/table).
+            (Defaults to 0.7)
 
     Returns:
         fig (matplotlib.figure.Figure): Closed Matplotlib Figure object
@@ -636,7 +649,10 @@ def multi_time_pie_grid(
     times = sorted(agg['time'].unique())
     N = len(times)
     nrows = int(np.ceil(N / ncols))
-    fig = plt.figure(figsize=(5 * ncols + 3, 5 * (nrows + 0.7)))
+    fig = plt.figure(figsize=(
+        individual_pie_dimension * ncols + horizontal_buffer,
+        individual_pie_dimension * (nrows + vertical_buffer)
+    ))
 
     # Create table gridspec
     left = 0.05


tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -209,6 +236,209 @@ def construct_legend(ax, data_comp=False):
...
+    # Create table gridspec
+    left = 0.05
+    top = 0.95
+    gs = fig.add_gridspec(
+        nrows + 1, ncols,
+        height_ratios=([1.0] + [1]*nrows),
+        hspace=0.4,
+        wspace=0.3,
+        left=left, right=(1-left), top=top, bottom=(1-top)
+    )
```
gonuke: Separation of concerns... I think the `GridSpec` should be generated at a different level, and then the specific `table_ax` passed into this method to populate it.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -209,6 +236,209 @@ def construct_legend(ax, data_comp=False):
...
+    col_labels = [f'{t} {time_unit}' for t in times]
+    row_labels = [reformat_isotope(nuc) for nuc in all_nucs]
+
+    table_data_float = np.array([
+        [float(cell.strip('%')) for cell in row] for row in table_data
+    ])
+    row_means = table_data_float.mean(axis=1)
+    sorted_indices = np.argsort(-row_means)
+    table_data = sort_by_index(table_data, sorted_indices)
+    row_labels = sort_by_index(row_labels, sorted_indices)
```
gonuke: Wouldn't all this sorting be easier if it was a dataframe still?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -209,6 +236,209 @@ def construct_legend(ax, data_comp=False):
...
+        left=left, right=(1-left), top=top, bottom=(1-top)
+    )
+
+    # Build and populate table
+    table_ax = fig.add_subplot(gs[0,:])
+    table_ax.axis('off')
+    table_data = populate_table_rows(agg, all_nucs, times)
+
+    col_labels = [f'{t} {time_unit}' for t in times]
+    row_labels = [reformat_isotope(nuc) for nuc in all_nucs]
```
gonuke: Can this formatting be done as it's passed in to the `table()` method?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -209,6 +236,209 @@ def construct_legend(ax, data_comp=False):
...
+
+def pie_grid_relative_tables(
+    fig, agg, all_nucs, times, nrows, ncols, time_unit, color_map
+):
+    '''
+    Given a Matplotlib Figure object, create a subplot Matplotlib table object
+        to act as a color-coordinated legend for a grid of pie charts with
+        the percentage contributions of each nuclide represented in any of the
+        pie charts at each cooling time (as well as a row for aggregated
+        "Other" nuclides taht fall below a given proportional threshold).
```
gonuke:
```suggestion
        "Other" nuclides that fall below a given proportional threshold).
```

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -341,4 +571,196 @@ def plot_single_response(
...
+        run_lbl=run_lbl,
+        variable=variable,
+        time_unit=time_unit,
+        threshold=threshold
+    )
+    all_nucs = sorted(set(agg['nuclide']))
+    color_map = build_color_map_from_nuc_list(all_nucs, cmap_name=cmap_name)
+
+    N = len(times)
+    nrows = math.ceil(N / ncols)
```
gonuke: Can you use `np.ceil` to avoid importing `math` for this one thing?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -341,4 +571,196 @@ def plot_single_response(
...
+        run_lbl (str): Distinguisher of the specified ALARA run.
+        variable (str): Name of the response variable.
+        threshold (float): Proportional threshold for small-value aggregation.
+        time_idx (int): Cooling time interval number (i.e. 0 for shutdown).
+        time_unit (str, optional): Optional paramter to set units for cooling
+            times. Accepted values: 's', 'm', 'h', 'd', 'w', 'y', 'c'.
+            (Defaults to 's')
+    
+    Returns:
+        None
```
gonuke: Should we return the `plt` so that the user can continue to manipulate it? and then maybe let them `show()` it when ready (separation of concerns)

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -209,6 +236,209 @@ def construct_legend(ax, data_comp=False):
...
+
+    values = time_slice['value'].to_numpy()
+    total = time_slice['value'].sum()
+    fractions = values / total if total > 0 else np.zeros_like(values)
+
+    labels = []
+    if np.any(fractions == 1.0):
+        for nuc, frac in zip(time_slice['nuclide'], fractions):
+            labels.append(reformat_isotope(nuc) if frac == 1.0 else '')
+        return labels
```
gonuke: Isn't this edge case already covered by the case below, as long as you handle `Other` more carefully? Maybe the condition is:
```
if frac >= threshold or (nuc == 'Other' && frac > 0):
```

gonuke (changes requested):
Nicely formatted and put together here... I have a number of suggestions to 
* help separate concerns a little and 
* take advantage of dataframes without creating so many new data structures

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -209,6 +236,209 @@ def construct_legend(ax, data_comp=False):
...
+        height_ratios=([1.0] + [1]*nrows),
+        hspace=0.4,
+        wspace=0.3,
+        left=left, right=(1-left), top=top, bottom=(1-top)
+    )
+
+    # Build and populate table
+    table_ax = fig.add_subplot(gs[0,:])
+    table_ax.axis('off')
+    table_data = populate_table_rows(agg, all_nucs, times)
```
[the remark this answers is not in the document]
eitan-weinstein: No reason other that I just didn't know about `pandas.plotting.table()` until now!

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -113,6 +107,30 @@ def build_color_map(list_of_pivs, cmap_name='Dark2'):
 
     return {lbl: cmap(i % cmap.N) for i, lbl in enumerate(sorted(all_nucs))}
 
+def build_color_map_from_nucs(all_nucs, cmap_name):
```
gonuke: It's not obvious how this is substantially different from `build_color_map_from_pivs`?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -340,5 +527,216 @@ def plot_single_response(
...
+    '''
+
+    color_map = build_color_map(
+        cmap_name=cmap_name, all_nucs=agg['nuclide'].unique()
+    )
+
+    times = sorted(agg['time'].unique())
+    N = len(times)
+    nrows = int(np.ceil(N / ncols))
+    fig = plt.figure(figsize=(5 * ncols + 3, 5 * (nrows + 0.7)))
```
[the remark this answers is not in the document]
gonuke: or give them some name that evokes some kind of meaning...
```
padding = 5
vspace = 3
hspace = 4.5
```

gonuke (approved):

PR #180, opened 2025-12-02
Change alara_output_processing.py to add colons as separator

anu1217:
Currently, `parse_table_data()` in alara_output_processing.py uses ` #` as a separator to identify the block type and block number. However, if the block type is a zone, there is a colon at the end of the zone number. This would require the user to specify block_num as `1:`, for example, when filtering output data with the alara_output_processing module. This PR introduces a change intended to prevent the colon from being a necessary addition to the value of block_num.

### DIFF
--- a/tools/alara_output_processing/alara_output_processing.py
+++ b/tools/alara_output_processing/alara_output_processing.py
@@ -3,7 +3,6 @@ import argparse
 from warnings import warn
 from csv import DictReader
 from numpy import array
-import re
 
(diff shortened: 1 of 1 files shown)


gonuke (changes requested):
Good catch @anu1217 - I wonder if we should make ALARA behave more consistently?

In the meantime, I think `re` may be a bigger hammer than we need here.

gonuke (approved):
Thanks @anu1217

PR #183, opened 2025-12-04
Adding block_name to canonical ALARADFrame structure

### LINKED ISSUE #182: Include block name in canonical ALARADFrame structure
Include ALARA postprocessing to read in the "Block" (Interval, Zone, Material) name in addition to its type and number in the canonical structure of ALARADFrames parsed from the files.

eitan-weinstein:
Closes #182.

This PR responds to @anu1217 's request to include block names in the canonical structure for ALARDFrames, in addition to `block` and `block_num`.


gonuke:
> If there is no name, `block_name` will just be an empty string. Is there an appropriate filler you have in mind or do you think it would be fine to blank (either way, the string operations can handle a missing name and do not cause errors)?

This was a lazy question that wondered what `partition` did under different circumstances and to make sure that particular line wasn't going to be a problem....  it looks like it will be fine.

Issue #187, opened 2025-12-05
a method to combine/amalgamate cross-sections that produce the same child nuclide (whether gases or otherwise)


gonuke:
In support of data processing options described in #184

Issue #186, opened 2025-12-05
a method to subtract the individual cross-sections that produce gases as child nuclides from the total gas production cross-section for the corresponding gas


gonuke:
In support of data processing errors described in #184

Issue #185, opened 2025-12-05
a method to remove all cross-section that have gases as child nuclides


gonuke:
In support of data processing errors defined in #184

Issue #184, opened 2025-12-05
Add methods to clean up data from NJOY processing


gonuke:
The ALARAJoy wrapper currently processes all reaction channels available after running a series of NJOY modules, including:
* separate reaction channels that produce the same child nuclide (whether gases or otherwise)
* reaction channels that produce child nuclides that are also produced gases
* total gas production cross sections for a number of specific gases 

This has a number of consequences:
* error: over-production of gases through both the channel(s) where the gas is the child nuclide and the channel of total gas production, which (presumably) includes that/those channels
* computational cost: two channels that produce the same child nuclide  (whether gases or otherwise) could be combined into a single channel documented as the amalgam of multiple channels to (a) avoid the cost of processing multiple channels and (b) to avoid premature truncation due to production spread across channels

To fix this, we should have a series of processing steps that will clean up the GENDF data dictionary after it has been loaded from the GENDF file
1. #185
2. #186
3. #187

One of the first two much be called to result in a correct data set!

The last one is optional and trades off between more fidelity in the reaction channels (assuming we trust that fidelity in the original data) and lower computational cost.

PR #190, opened 2025-12-05
Saving all gas production reaction pathways

eitan-weinstein:
Primer for #184.

This PR restructures the gas reaction handling for `ALARAJOYWrapper` to simply write out all reaction pathways -- including gas residuals -- to memory in `gendf_data`. From here, the subsequent methods that I will develop for #185, #186, and #187 can be applied to `gendf_data`.

Additionally, I noticed when going through these data pathways the total alpha production was being cut out because of an improper `range` setting in `preprocess_fendl3.py`, which was cutting off the MT union between the gases and the PENDF MTs at MT = 206, instead of 207, so I fixed that in this PR as well.

### DIFF
--- a/tools/ALARAJOYWrapper/reaction_data.py
+++ b/tools/ALARAJOYWrapper/reaction_data.py
@@ -5,7 +5,8 @@ from numpy import array
 # changes in a nucleus following neutron activation
 
 #             emitted         delta N       delta P
-NP_dict   = {'n'     : array([-1      ,      0      ]), # neutron emission
+NP_dict = {
+             'n'     : array([-1      ,      0      ]), # neutron emission
              'g'     : array([ 0      ,      0      ]), # gamma emission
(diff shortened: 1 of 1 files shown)


tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -81,9 +81,8 @@ def emission_breakdown(emitted_particles):
     }
 
     # Handle gas production totals (MTs 203-207) of the format Xz
-    gases = ['p', 'd', 't', 'h', 'a']
     if 'X' in emitted_particles:
-        for gas in gases:
+        for gas in list(NP_dict.keys())[2:]:
```
gonuke: Readability: probably useful to still define
```
gases = list(NP_dict.keys())[2:]
```
maybe immediately following the definition of `NP_dict`

PR #191, opened 2025-12-08
Gas Handling Option #1: Remove Non-Total Gas Daughters

### LINKED ISSUE #185: a method to remove all cross-section that have gases as child nuclides

eitan-weinstein:
Closes #185.

This PR introduces the framework for users to determine their preferred gas handling method in `ALARAJOYWrapper`, as laid out in #184. This includes an additional required argument `-g` to set the desired gas handling method and a new filtering function `gas_handling()` to be called before  `write_dsv()`. This function directs to one of the two gas handling options, each of which will return a filtered list from `cumulative_data` of reactions satisfying that particular gas handling method.

Specifically, the method for removing reactions that produce gas daughters (excluding the MT = 203 - 207 gas total production cross-section "reactions"), as set by writing `-g r` in the command line arguments (for "remove"), is introduced and integrated into `gas_handling()`. My next PR will introduce the subtraction method described in #186.

### DIFF
--- a/tools/ALARAJOYWrapper/preprocess_fendl3.py
+++ b/tools/ALARAJOYWrapper/preprocess_fendl3.py
@@ -45,7 +45,24 @@ def amalgamate_by(nuc_type, row_dicts):
 
     return amalgam
 
-def remove_gas_daughters(cumulative_data):
+def build_parent_daughter_struct(cumulative_data):
+    by_parent = defaultdict(lambda: defaultdict(dict))
+
+    for rxn in cumulative_data:
+        parent = rxn['Parent KZA']
+        daughter = rxn['Daughter KZA']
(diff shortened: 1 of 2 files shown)


gonuke (changes requested):
This now filters out all gas daughters, even if there is only a single gas production reaction.

I'm increasingly convinced that we might want a data structure that is either a dataframe or
```
parentKZA :
   daughterKZA :
      MT :
         emitted : string
         xsection : numpy array
```
This means the data is always amalgamated by parent and easy to reference by daughter.

Then, this function could be:
```
for each parent:
    for each gas:
       if there is a gas production xsection for that gas:
          delete other reactions with that gas daughter
```

PR #192, opened 2025-12-08
Gas Handling Option #2: Subtract Gas Cross-Sections from Totals

### LINKED ISSUE #186: a method to subtract the individual cross-sections that produce gases as child nuclides from the total gas production cross-section for the corresponding gas

eitan-weinstein:
Closes #186.

This PR is a follow up to #191, offering the alternative method for gas handling within `ALARAJOYWrapper`. Here, instead of removing reactions that produce a gas daughter, those reactions' cross-sections will be calculated by subtraction from the gas production totals corresponding to that specific gas daughter.

### DIFF
--- a/tools/ALARAJOYWrapper/preprocess_fendl3.py
+++ b/tools/ALARAJOYWrapper/preprocess_fendl3.py
@@ -93,11 +93,12 @@ def subtract_gas_from_totals(all_rxns, gas_tuples):
 
     for parent in all_rxns:
         for gKZA, gMT in gas_tuples:
-            mt_data = all_rxns[parent][gKZA]
-            if gKZA in all_rxns[parent] and gMT in mt_data:
-                for gRxn in mt_data:
+            if gKZA in all_rxns[parent] and gMT in all_rxns[parent][gKZA]:
+                for gRxn in all_rxns[parent][gKZA]:
                     if gRxn != gMT:
-                        mt_data[gMT]['xsections'] -= mt_data[gRxn]['xsections']
+                        all_rxns[parent][gKZA][gMT]['xsections'] -= (
+                            all_rxns[parent][gKZA][gRxn]['xsections']
+                        )
 
     return all_rxns
 
@@ -133,6 +134,7 @@ def gas_handling(gas_method, all_rxns):
     """
 
     gas_tuples = list(rxd.GAS_DF[['kza', 'total_mt']].itertuples(index=False))
+
     if gas_method == 'r':
         return remove_gas_daughters(all_rxns, gas_tuples)
 
@@ -140,8 +142,8 @@ def gas_handling(gas_method, all_rxns):
          return subtract_gas_from_totals(all_rxns, gas_tuples)
     
     else:
-        raise KeyError(
-            'Invalid gas method key. ' \
+        raise ValueError(
+            'Invalid gas method selection. ' \
             'Must choose either "r" (remove) or "s" (subtract).'
         )
 


gonuke (changes requested):
An over all algorithm simplicity comment - happy to discuss

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -47,18 +49,85 @@ def remove_gas_daughters(all_rxns):
...
+        all_rxns (collections.defaultdict): Modified version of all_rxns with
+            each reaction that produces a gas daughter calculating its cross-
+            sections through subtraction from gas production totals.
+
+    """
+
+    gas_tuples = list(rxd.GAS_DF.itertuples(index=False))
+    for parent in all_rxns:
+        gas_accum = defaultdict(lambda: None)
+        for daughter in all_rxns[parent]:
```
gonuke: I don't think we should ever have to loop over parents & daughters.  We should only every have to loop over parents & gases, and then occassionally over MT for a given gas daughter.

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -47,18 +49,85 @@ def remove_gas_daughters(all_rxns):
...
+                }    
+            }
+    
+    Returns:
+        all_rxns (collections.defaultdict): Modified version of all_rxns with
+            each reaction that produces a gas daughter calculating its cross-
+            sections through subtraction from gas production totals.
+
+    """
+
```
gonuke: This method looks pretty complicated.  Before I review it closely, I was thinking we would have something like this pseudo code:
```
for each parent in all_rxns:
   for each gasKZA, gasMT in gas_data:
      if there is a gasKZA among the daughters and a gasMT among those reactions
         for each gasRxn in all_rxns[parent][gasKZA]:
            if gasRxn != gasMT:
               mult = getGasProductionMultiplier(gasKZA,gasRxn)
               all_rxns[parent][gasKZA][gasMT] -= all_rxns[parent][gasKZA][gasRxn] * mult
```

gonuke (changes requested):
More comments on algorithm simplification

tools/ALARAJOYWrapper/preprocess_fendl3.py
eitan-weinstein: You're absolutely right. I was thinking about this the wrong way, assuming that this subtraction would necessarily eliminate all 'x' emission reactions, when in reality, this is not the case because there could be some products that would only be accounted for as emitted particles, and would be inappropriately deleted in my version.

That said, I'm not sure I follow what your intended goal is for the `getGasProductionMultiplier()` function that you show above. Could you elaborate on why we need a multiplier?

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -45,20 +45,62 @@ def remove_gas_daughters(all_rxns):
...
+    for parent in all_rxns:
+        for gKZA, gMT in gas_tuples:
+            mt_data = all_rxns[parent][gKZA]
```
gonuke: We don't know if `gKZA` is in `all_rxns[parent]` yet - we test that on the next line, so might be easiest to use the full variable like we do in `remove_gas_daughters`

gonuke (approved):
This look clean now @eitan-weinstein - thanks

PR #195, opened 2025-12-09
Comparison Tool 5.5: Filtering out Stable Nuclides of Irradiated Element

### LINKED ISSUE #194: Filter Out Stable Nuclides of an Irradiated Element in Single Element Simulation
Per discussions with @bohmt, we determined that it would be helpful to filter out all of the stable nuclides of an irradiated target to get a clearer picture of the transmutation products at shutdown. In particular, this feature can be used within the function `tools/alara_output_processing/alara_output_plotting.single_time_pie_chart()` to zoom in on the number density of small transmutation products at shutdown, that may still contribute significantly to the post-irradiation analysis of the sample that can be compared across data sources.

eitan-weinstein:
Closes #194.

I created a new issue within the FENDL3.x Validation Milestone for this PR that details the motivation for this new feature more specifically, but essentially, the purpose of the PR is to allow the user to ignore the stable isotopes of the irradiated target to see only *new* nuclides produced as a result of the reaction, whether stable or not.


gonuke:
I'm not thrilled about hard-coding the list of stable nuclides.  We should be able to infer this from various data libraries, but the best solution may be to require/expect the specific activity to be generated in the ALARA run and use those results to filter the number density.  (That is, if the user doesn't just look at the specific activity results, anyway.)

I guess a question for @bohmt - the motivation is to remove the stable nuclides from overwhelming the result because they aren't important for other responses, but shouldn't users just look at those other responses?

PR #197, opened 2025-12-10
Creating Hierarchical Data Structure for ALARAJOYWrapper

eitan-weinstein:
Closes #196 .

This PR changes the structure in which reaction data are stored in `ALARAJOYWrapper` from a list of dictionaries to a hierarchical multi-nested dictionary keyed by parent nuclides, of the form:

```
{parent:
    {daughter:
        {MT:
            {
                Emitted Particles:
                Non-Zero Groups:
                Cross Sections:
            }
        }
    }    
}
```

A number of corresponding changes were made to accommodate this new structure.

PR #198, opened 2025-12-10
Gas Handling Option #1a: Remove Non-Total Gas Daughters (Following Data Restructuring)

eitan-weinstein:
Supplants #191.

I made this into a new PR given the extent of the changes to the data structure in `ALARAJOYWrapper` made the old PR so outdated that it might be better to start a new one. Additionally, I figured this would be a good opportunity to try following @gonuke 's request that I create new feature PRs into my own fork if they are dependent on other PRs in review (in this case #197).

Original text of PR description:

> This PR introduces the framework for users to determine their preferred gas handling method in ALARAJOYWrapper, as laid out in https://github.com/svalinn/ALARA/issues/184. This includes an additional required argument -g to set the desired gas handling method and a new filtering function gas_handling() to be called before write_dsv(). This function directs to one of the two gas handling options, each of which will return a filtered list from cumulative_data of reactions satisfying that particular gas handling method.

> Specifically, the method for removing reactions that produce gas daughters (excluding the MT = 203 - 207 gas total production cross-section "reactions"), as set by writing -g r in the command line arguments (for "remove"), is introduced and integrated into gas_handling(). My next PR will introduce the subtraction method described in https://github.com/svalinn/ALARA/issues/186.

### DIFF
--- a/tools/ALARAJOYWrapper/preprocess_fendl3.py
+++ b/tools/ALARAJOYWrapper/preprocess_fendl3.py
@@ -51,8 +51,8 @@ def remove_gas_daughters(all_rxns):
             with double-counted gas-producing reactions left out.
     """
 
+    gas_tuples = list(tp.GAS_DF[['kza', 'total_mt']].itertuples(index=False))
     for parent in all_rxns:
-        gas_tuples = tp.GAS_DF[['kza', 'total_mt']].itertuples(index=False)
         for gKZA, gMT in gas_tuples:
             if gKZA in all_rxns[parent] and gMT in all_rxns[parent][gKZA]:
                 all_rxns[parent][gKZA] = {gMT: all_rxns[parent][gKZA][gMT]}


tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -2,81 +2,178 @@
...
+            double-counted gas-producing reactions left out.
+    """
+
+    for parent in all_rxns:
+        for daughter in all_rxns[parent]:
+            MTs = set(all_rxns[parent][daughter])
+            for MT in MTs:
+                gas_MTs = tp.GAS_DF['total_mt'].tolist()
+                if MT not in gas_MTs and any(MT in gas_MTs for MT in MTs):
+                    del all_rxns[parent][daughter][MT]
```
gonuke: I think I'd loop this differently  - something like this.
```suggestion
    gas_data = tp.GAS_DF['total_mt'].tolist()
    for parent, rxns in all_rxns.items():
        daughterKZAs = list(rxns.keys())
        for gasKZA, gas_MT in tp.GAS_DF['kza','total_mt']:
            if gasKZA in daughterKZAs and gas_MT in rxns[gasKZA].keys():
               # delete all non gas_MTs
```

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -13,12 +13,84 @@ def args():
...
+            }
+    
+    Returns:
+        all_rxns (collections.defaultdict): Modified version of all_rxns with
+            double-counted gas-producing reactions left out.
+    """
+
+    gas_MTs = set(tp.GAS_DF['total_mt'])
+    for parent in all_rxns:
+        for daughter in all_rxns[parent]:
```
gonuke: Since we have many more daughters than gases, it is much more efficient to loop over the gases here and not all the daughters.

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -13,12 +13,84 @@ def args():
...
+            MTs = list(all_rxns[parent][daughter])
+            if any(MT in gas_MTs for MT in MTs):
+                for MT in MTs:
+                    if MT not in gas_MTs:
+                        del all_rxns[parent][daughter][MT]
```
gonuke: set operations could make this cleaner
```suggestion
            MTs = set(all_rxns[parent][daughter])
            if MTs.intersection(gas_MTs):
                for (MT in MTs - gas_MTs):
                    del all_rxns[parent][daughter][MT]
```

gonuke (changes requested):
Thanks for figuring out the problem with `itertuples`... I have a question about ways that we can move this out of the loop, though, since it seems wasteful and slightly confusing to regenerate this constant data with each `parent`

PR #213, opened 2025-12-22
Option to Avoid Ground State Isomer Decay Assumption

### LINKED ISSUE #212: Amalgamate Isomers without Cross-Sections to "Other" Destruction Cross-Section
Alternate option to #207 . Instead of assuming that all excited states of an isotope will necessarily decay to the ground state of that same isotope, create a new "Other" daughter for each parent. Will need to define some appropriate KZA value that will distinguish this new pseudo-daughter that can be processed still by ALARA. 

Such an amalgamation can be helpful in determining the relative significance of these cross-sections by parent.

eitan-weinstein:
Closes #212.

This PR introduces an optional alternate pathway to avoid making the physics assumption that all ultra-short-lived isomers lacking known half-lives will necessarily decay to their ground states. Rather, the methods introduced in this PR create a new reaction row with a pseudo-daughter for all of these excitation reactions signified with a `*` as the KZA isomer tag (i.e. `26054*` for all relevant (n,n) excitation reactions for <sup>54</sup>Fe). All of these cross-sections are accumulated to this row instead of the ground state.

### DIFF
--- a/tools/ALARAJOYWrapper/README.md
+++ b/tools/ALARAJOYWrapper/README.md
@@ -32,18 +32,13 @@ ALARAJOYWrapper is designed to produce a space-delimited DSV containing data tha
 To run this preprocessor, the user must first have acquired TENDL files for each isotope to be processed from [TENDL 2017](https://tendl.web.psi.ch/tendl_2017/tendl2017.html), which is the source for FENDL3.2x neutron activation data. All TENDL files to be processed must be in the same directory as each other to be properly identified by ALARAJOYWrapper.
 
 Running ALARAJOYWrapper can be done with one Python command:
-
-    python preprocess_fendl3.py -g gas_handling_method -f /path/to/fendl3_data_dir/ -d /path/to/eaf_decay_library/ -a
-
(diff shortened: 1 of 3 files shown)


tools/ALARAJOYWrapper/README.md
```diff
@@ -33,13 +33,18 @@ To run this preprocessor, the user must first have acquired TENDL files for each
...
+This command takes five arguments,`-g `, `-f`, `-d`, `-i`, and `-a`. Of these, `-g`, `-f`, and `-d` are required. The `-g` argument sets the gas handling method for gas totals, as calculated by the NJOY GASPR module. The two possible methods are `r` (remove) and `s` (subtract). Method `r` removes all reactions with a light gas daughter (protons through alpha particles), with the exception of total gas production cross-sections corresponding to MT = 203-207 reactions (see Appendix B of the ENDF-6 Manual at `../../developer-info/endf6-manual.pdf`) when a total gas production cross-section exists. Method `s` subtracts the cross-sections for each energy group of reactions that produce a gas daughter from the total gas production cross-sections instead. In effect, for gas producing reactions, selecting `r` will show only the *total* gas production cross sections for each gas, whereas `s` will show each *individual* gas production pathway with its respective cross section data.
```
gonuke: As this gets more complicated, I wonder if it's better to just make the `parseargs` help messages more robust and have users get the documentation there.  It will also help indicate which things are required and which are optional

gonuke:
I'm waiting for #209 to be merged because this has overlapping code...

Issue #206, opened 2025-12-18
Add a column to the ALARA dataframe showing the half-life for the unstable isotopes


bohmt:
When looking at tables of number densities, it is useful for activation analysts to quickly see if a given nuclide is stable or unstable. Further if unstable, it is useful to see the half-life of that nuclide.
See the attached snippet of the output from FISPACT-II.

[testinventoryFe2yr.txt](https://github.com/user-attachments/files/24245919/testinventoryFe2yr.txt)

PR #217, opened 2025-12-23
ALARADFrame Half-Life Filtering

### LINKED ISSUE #202: Add a method to identify the radioactive nuclides in an output data set
This will require users to have included `Specific Activity` in their output and then becomes simply the process of extracting the nuclides from that ALARA DataFrame since they are all radioactive.

### LINKED ISSUE #203: Add a method to filter another ADF with the radioactive nuclide list
This may not need to be a method since I think it's a one-liner based on your filtering methods.

### LINKED ISSUE #216: Create method to filter based on magnitude of half-lives
Following #215, filtering stable or unstable nuclides becomes a very simple check of whether a nuclide has a half-life or is stable. However, with this tweak, it also is fairly straightforward to allow for time based filtering of half-lives above or below a certain threshold.

eitan-weinstein:
Closes #216.
Closes #203.
Closes #202 .

This PR follows the inclusion of half-lives into the canonical structure of ALARADFrames in #206. I've introduced filtering capabilities in `alara_output_processing.filter_rows()` for users to filter on the stability or lack thereof of the nuclides in the ALARADFrame, as well as filtering half-lives against time thresholds (i.e. preserving all nuclides with a half-live greater than 1e6 seconds).

### DIFF
--- a/tools/alara_output_processing/alara_output_processing.py
+++ b/tools/alara_output_processing/alara_output_processing.py
@@ -1,20 +1,13 @@
 import pandas as pd
 import argparse
-import operator
+from operator import gt, lt
 from warnings import warn
 from csv import DictReader
 from numpy import array
 
 # ---------- General Utility Methods ----------
 
-OPS = {
-    '>'  : operator.gt,
-    '<'  : operator.lt,
-    '>=' : operator.ge,
-    '<=' : operator.le,
-    '==' : operator.eq,
-    '!=' : operator.ne
-}
+OPS = { '>' : gt, '<' : lt }
 
 SECONDS_CONV = {'s': 1}
 SECONDS_CONV['m'] = 60  * SECONDS_CONV['s']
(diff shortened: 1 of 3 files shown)


tools/alara_output_processing/alara_output_processing.py
```diff
@@ -359,15 +369,85 @@ def _single_element_all_nuclides(self, element):
...
+        
+        Returns:
+            half_lives (set): Set of all numeric half-lives in the ALARADFrame
+                (i.e. excluding "stable" for stable nuclides).
+        '''
+
+        half_lives =  set(self.loc[
+            self['half_life'].apply(lambda x: isinstance(x, float)),
+            'half_life'
+        ])
```
gonuke: While this isn't too complicated, it makes we wonder if it isn't better to use some special float to indicate stable nuclides, e.g. -1?  Then this could just be  a simpler filter for `half_life >= 0`?

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -1,11 +1,21 @@
...
 
 # ---------- General Utility Methods ----------
 
+OPS = {
+    '>'  : operator.gt,
+    '<'  : operator.lt,
+    '>=' : operator.ge,
+    '<=' : operator.le,
+    '==' : operator.eq,
+    '!=' : operator.ne
```
gonuke: Given the precision of floating point operations I wonder if we need any of these that include equality? Limiting it to just `<` and `>` might make it all a little simpler, too.

gonuke (changes requested):
Perhaps we should discuss as a group (incl. via slack) whether we need any half-life operators that include equality.  Since this is float data, it is always dubious to use any real equality comparison anyway.

PR #224, opened 2026-01-06
Adding missing return line in rxn_to_str()

eitan-weinstein:
I realized that the new function `rxn_to_str()` in `preprocess_fendl3.py` was missing its return line, so the DSV writing was failing (unfortunately only _after_ trying to go through a full run with FENDL3!).


gonuke (approved):
Looks strightforward.

PR #228, opened 2026-01-15
Write pulsing history details to output

### LINKED ISSUE #218: Develop a function to always write pulse history details
`ScheduleItem` has a `write()` method that always shares information about a schedule to the output file, independent of verbosity, but it only refers to the pulse history by name and doesn't include any details.

This issue will closed when a new `write()` method exists for `PulseHistory()` that can be called to also put pulse history details in the output file.  Some details are shared with verbosity > 3 already, but in the wrong place and not a useful format.

The formatting should focus machine readability rather than human readability, although one method could be written for each case.

anu1217:
This PR modifies PulseHistory() to take the name of each pulsing history and print it, along with the number of levels in the history, number of pulses in each level, and delay time in seconds to the output file. This information is always printed to the output file, regardless of verbosity and whether the pulsing history appears in any schedule.

fixes #218

### DIFF
--- a/src/PulseHistory.C
+++ b/src/PulseHistory.C
@@ -233,11 +233,10 @@ void PulseHistory::write_ph() const
     }
     cout << "]" << endl;
 
-    cout << "delay_seconds_per_level: [";
-    for (int lvlNum = 0; lvlNum < nLevels; lvlNum++)
+    cout << "delay_seconds_per_level: [" << td[0];
+    for (int lvlNum = 1; lvlNum < nLevels; lvlNum++)
     {
-      cout << td[lvlNum];
-      if (lvlNum < nLevels - 1) cout << ", ";
+      cout << ", " << td[lvlNum];
     }
     cout << "]" << endl;
 }


src/PulseHistory.C
```diff
@@ -206,3 +222,24 @@ Matrix PulseHistory::doHistory(Matrix opT)
...
+      if (lvlNum < nLevels - 1) cout << ", ";
+    }
+    cout << "]" << endl;
+
+    cout << "delay_seconds_per_level: [";
+    for (int lvlNum = 0; lvlNum < nLevels; lvlNum++)
+    {
+      cout << td[lvlNum];
+      if (lvlNum < nLevels - 1) cout << ", ";
+    }
```
gonuke: Same as above:

```suggestion
    cout << "delay_seconds_per_level: [" << td[0];
    for (int lvlNum = 1; lvlNum < nLevels; lvlNum++)
    {
      cout << ", " << td[lvlNum];
    }
```

PR #229, opened 2026-01-20
Handling isomers produced beyond (n,n*) reactions

eitan-weinstein:
This PR expands on the isomer handling methods for ALARAJOYWrapper developed over the past month to properly address isomer daughters produced from pathways other than (n,n*) reactions. I came across this issue when doing inventory benchmarking against FISPACT-II and the JAEA FNS experimental data and noticed certain isomers appearing to be stable when they obviously were not, just like before. The old method failed to include cases where the parent itself is an isomer, which can subsequently produce excited daughters by normal reaction pathways.


tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -271,21 +273,22 @@ def iterate_MTs(MTs, file_obj, mt_dict, pKZA, all_rxns, radionucs, to_ground):
         # or any ground-state daughters. Necessarily need to cut off maximum
         # excitation at 9th state by nature of KZA conventions
         if (
-            _is_ground_state_nuc(rxn['isomer']) or
-            _is_isomer_with_decay_data(dKZA, radionucs, rxn['isomer'])
+            _is_ground_state_daughter(rxn['isomer'], dKZA) or
+            _is_isomer_with_decay_data(dKZA, pKZA, radionucs, rxn['isomer'])
```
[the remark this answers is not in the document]
gonuke: `_is_isomer_with_decay_data` can just be refactored to be `_has_decay_data` and doesn't need to check if it's an isomer

tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -269,34 +292,34 @@ def iterate_MTs(MTs, file_obj, mt_dict, pKZA, all_rxns, radionucs, to_ground):
```
gonuke: Can we "fix" `dkza` here to make it "valid" right here? Then we won't need `_has_valid_daighter`

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -118,8 +118,61 @@ def set_directory():
...
+
+    za = str((pKZA + mt_dict[MT]['delKZA']) // 10).zfill(6)
+    for MF in isomer_dict[MT].keys():
+        if len(isomer_dict[MT][MF]) > 1:
+            for M in isomer_dict[MT][MF]:
+                card9_lines.append(
+                    f'-1 / \n {MF} {MT} {za} {M} "{mtname}, M={M}" /'
+                )
+
+    return card9_lines
```
gonuke: If we pass in `mt_dict[MT]` as `mt_data` and `isomer_dict[MT]` as `isomer_data`
and return a fresh set of `card9_lines` to be appended by the function that calls this one.

```suggestion
    card9_lines = []
    za = str((pKZA + mt_data['delKZA']) // 10).zfill(6)
    for MF in isomer_data:
        for M in isomer_data[MF]:
            card9_lines.append(
                f'-1 / \n {MF} {MT} {za} {M} "{mtname}, M={M}" /'
            )

    return card9_lines
```

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -158,9 +217,16 @@ def fill_input_template(
 
     card9_lines = []
     MFD = 3 # ENDF file tag for cross-section data
-    for MT in MTs:
-        mtname = mt_dict[MT]['reaction']
-        card9_lines.append(f'{MFD} {MT} "{mtname}" /') 
+    if isomer_dict:
```
gonuke: Based on the docstring, every time `isomer_dict` is present means that `pKZA` should also be present.  Is that right?  Or vice versa?

Later, after you check for `isomer_dict` you also check for `pKZA`.  The logic is a little opaque...

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -310,60 +447,24 @@ def main():
         endf_path = file_properties['TENDL File Path']
         TAPE20.write_bytes(endf_path.read_bytes())
 
-        material_id, MTs, endftk_file_obj = tp.extract_endf_specs(TAPE20)
+        material_id, MTs = tp.extract_endf_specs(TAPE20)
         MTs = set(MTs).intersection(mt_dict.keys())
```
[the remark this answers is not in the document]
gonuke: Let's add an issue for this.  I don't think it should hold this up, but I do think we should report MTs that end up in the TENDL file that are not in our `mt_dict` that are outside some set of expected exclusions.

tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -75,158 +118,248 @@ def extract_endf_specs(path):
...
+            that can be referenced with known-decay data from the provided EAF
+            decay library.
     """
 
-    return dKZA in radionucs and M in range(1,10)
+    trial_M = min(M-1, 9)
+    while (dKZA + trial_M) not in eaf_nucs and trial_M >= 0:
+        trial_M -= 1
+
+    return dKZA + trial_M
```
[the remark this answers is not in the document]
gonuke: I do see your later comment about this. Maybe I still don't have all the nuances of our logic about how to handle these cases in my head.  I wonder if you can add a simple pseudo-code logic to the readme about how we are handling these?

I think having the loop exit with `trial_M == -1` is a problem because `dKZA` will end up being reduced by 1 which will perturb the downstream logic.  I'll let you propose a way to reconcile this.

tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -75,158 +118,248 @@ def extract_endf_specs(path):
...
     Returns:
-        sigma_list (list): All of the cross-sections for a given reaction type
-            and material, listed as floating point numbers. If the run fails,
-            the function will just return an empty list.
+        MTs (set): Set of all unique MTs contained in the GENDF file.    
     """
 
-    section = file.section(MT).content
+    MTs = set()
+    with open(gendf_path, 'r') as f:
```
gonuke: You now parse each file twice, once just to read the MTs and accumulate all that are != 0, and then you pass that list to `extract_gendf_xs_line()` to read the x-sections.

Couldn't you combine these two methods and pass back the list of MTs from `extract_gendf_xs_lines`?

Issue #205, opened 2025-12-18
Create tool to visualize irradiation schedules


gonuke:
ALARA allows arbitrarily complex irradiation schedules.  It would be useful to be able to visualize those schedules.  An example of such a visualization is available in Wilson's PhD dissertation on pg 30.

PR #238, opened 2026-01-29
Create tool to plot simple irradiation histories

anu1217:
This tool plots irradiation histories from an ALARA output file using a bar chart/square wave-like representation. ~~Currently, it works for a simulation containing a single/top schedule with an arbitrary number of levels and an arbitrary number of pulse history blocks. In order to facilitate discussion of the tool's design and structure thus far, I decided to incorporate sub-schedules in a different PR.~~

This tool is currently able to plot an irradiation history consisting of a single-level schedule and single-level pulse history. I'd like to see how well this design works for this purpose before adding capabilities to process & plot more complex histories.

addresses #205

### DIFF
--- a/tools/plot_simple_sched_ph.py
+++ b/tools/plot_simple_sched_ph.py
@@ -66,7 +66,6 @@ def parse_arg():
     parser.add_argument("-f",
                         type=str,
                         required=True,
-                        default="/filespace/a/asrajendra/research/test_dir/simple_test_out",
                         help="path to file containing ALARA output")
     arg = parser.parse_args()
     return arg.f


gonuke:
This appears to have some overlap with #240... I think it makes sense to resolve #240 first and then add the plotting... unless I'm missing something?

anu1217:
I've added plots using two of the sample problems. Also made some changes to the post processor in order to handle verbose output that might appear between the schedule and pulse history details.

gonuke:
What does this look like with a plot history of 500 pulses of duration 5 days with a dwell time of  2 days?

PR #240, opened 2026-02-03
Store schedule and pulse history details from output for post processing

### LINKED ISSUE #220: Develop post-processing methods to read schedule into convenient data structure
Once the full details of an irradiation history are included in an ALARA input file, we can add methods to the post-processing tools to read that information into memory for usage by others.

anu1217:
This script writes the schedule and pulsing history related output to dictionaries so that they can be accessed and manipulated in other scripts. The schedule details are written in a nested dictionary to accommodate sub-schedules.

fixes #220

### DIFF
--- a/tools/sched_post_processor.py
+++ b/tools/sched_post_processor.py
@@ -97,8 +97,6 @@ def make_nested_dict(lines):
             ancestors.append(current_sched)
             current_sched = current_sched["children"][-1]
 
-        elif tokens[0] == "top_schedule":
-            ancestors.append(current_sched)
 
         elif tokens[0] == "pulse_entry:":
             current_sched["children"].append(make_pe_sub_dict(tokens))


tools/sched_post_processor.py
```diff
@@ -0,0 +1,124 @@
+import numpy as np
+import argparse
+
+def read_out(output_path):
+    with open(output_path, "r") as output_file:
+        lines = output_file.readlines()
```
gonuke: Just learned that this might make life easier by stripping all the lines when you read:
```suggestion
        lines = output_file.read().striplines()
```

tools/sched_post_processor.py
```diff
@@ -0,0 +1,124 @@
...
+    for line in lines:
+        if line.startswith("pulse_history:"): #next section of output
+            break
+
+        newline_name = line.lstrip('\t').rstrip('\n')
+        child_level = len(line.rstrip('\n')) - len(newline_name) # find number of tabs in the line
+
+        if newline_name.strip().split()[0] == 'schedule':
+            counter = 1
+            last_upper_indent_level[child_level][f'schedule {line.strip().split()[1]}'] = {
```
gonuke: You are using `new_line` in the conditional, and then `line` here.  I'm sure it works fine, but it would be cleaner if you used the same in both places.

tools/sched_post_processor.py
```diff
@@ -0,0 +1,124 @@
...
+    '''
+    sch_dict = {}
+    last_upper_indent_level = {0: sch_dict}
+    for line in lines:
+        if line.startswith("pulse_history:"): #next section of output
+            break
+
+        newline_name = line.lstrip('\t').rstrip('\n')
+        child_level = len(line.rstrip('\n')) - len(newline_name) # find number of tabs in the line
+
```
gonuke: it seems like you should `split()` this line always before the conditional and then avoid having to call split over and over again
```
    tokens = newline_name.split()
```

gonuke (changes requested):
There is a repeated suggestion here to use a while loop instead of a for loop when looking through the lines.  This is a typical design pattern for a parser.

A "design pattern" is typical and usually well-tested algorithm pattern for accomplishing a given task. Design patterns transcend languages and when you use them, others can quickly grasp what the intent of the code is and not worry about understanding the high level algorithm, but quickly focus on the details.

A "parser" is a typical pattern for moving through a stream of tokens (lines in your case) with two primary things that can happen at each line:
* change the "state"
* take action depending on the state

Zero, one, or both of these things can happen at each new token.

The most common action is to process the value of a token into some other data structure.

tools/sched_post_processor.py
```diff
@@ -0,0 +1,124 @@
...
+    return sch_dict
+
+def search_for_match(top_dict, search_str):
+    '''
+    Takes a nested top dictionary and searches all sub-dictionaries for matches with a provided string.
+    '''
+    matches = []
+    for key, value in top_dict.items():
+
+        if key.startswith(search_str):
```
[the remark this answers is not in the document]
anu1217: This is because of the `schedule ...` lines. Because the name of the schedule is not known beforehand, it is not possible to find an exact match. At first, I set each key to be just "schedule" without the name, but in that case make_nested_dict() will only store the last schedule line.

tools/sched_post_processor.py
```diff
@@ -0,0 +1,124 @@
...
+    'y' : 60 * 60 * 24 * 365,
+    'w' : 60 * 60 * 24 * 7,
+    'd' : 60 * 60 * 24,
+    'h' : 60 * 60,
+    'm' : 60,
+    's' : 1
+                     }
+    search_fields = ["schedule", "pulse_entry:"]
+    for search_field in search_fields:
+        matches = search_for_match(sch_dict, search_field)
```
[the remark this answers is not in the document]
anu1217: Initially I set this up such that the list was essentially a dummy list with only one entry, but this was not correctly capturing all of the entries. The list now contains multiple entries, which are values corresponding to each of the search field keys. The loop also iterates over each of the entries in the list.

tools/sched_post_processor.py
```diff
@@ -0,0 +1,126 @@
...
+            num_pulse_line = lines[line_idx+2].strip()
+            delay_line = lines[line_idx+3].strip()
+
+            pulse_hist_name = line.split()[1].strip("':")
+            num_pulses = num_pulse_line.split(":")[1]
+            delays = delay_line.split(":")[1]
+
+            pulse_dict[pulse_hist_name] = {
+                "num_pulses_all_levels" : num_pulses,
+                "delay_seconds_all_levels" : delays
```
gonuke: Do you want to convert these from strings to numbers here?
```suggestion
                "num_pulses_all_levels" : int(num_pulses),
                "delay_seconds_all_levels" : [float(d) for d in delays.split()]
```

tools/sched_post_processor.py
```diff
@@ -0,0 +1,126 @@
...
+            line_idx += 1 
+            new_child_level = lines[line_idx].count('\t')
+            counter = counter_dict[new_child_level] = 1
+
+        elif newline_name[0] == 'pulse_entry:':
+            counter = counter_dict[child_level]
+            last_upper_indent_level[child_level][f"pulse_entry num_{counter}_in_sched"] = make_pe_sub_dict(newline_name,
+                                                                                                           unit_multipliers)  
+            last_upper_indent_level[child_level + 1] = last_upper_indent_level[child_level][f"pulse_entry num_{counter}_in_sched"]
+            counter = counter_dict[child_level] = counter + 1
```
gonuke: I don't think you need to set `counter` here... you never use it before it gets reassigned

tools/sched_post_processor.py
```diff
@@ -0,0 +1,132 @@
...
+            current_sched = ancestors.pop()
+
+        if tokens[0] == "schedule":
+
+            current_sched["children"].append(make_sch_sub_dict(tokens))
+            ancestors.append(current_sched)
+            current_sched = current_sched["children"][-1]
+
+        elif tokens[0] == "top_schedule":
+            ancestors.append(current_sched)
```
gonuke:
```suggestion
```

PR #246, opened 2026-02-19
Small Updates to alara_output_plotting

eitan-weinstein:
As I've been going through looking at the longer term activation responses for the ALARAJOY-processed FENDL3 data and comparing it with FISPACT-II data, there have been a number of small changes that I've found beneficial to include into the ALARA output plotting module:

- Allowing the user to choose between plotting with `plt.plot()` or `plt.scatter()` when calling `plot_single_response()`
- Creating an option for a minimum y-axis cutoff (I'm seeing that responses will be going down to exceedingly small values (E-32) which is both beyond any actual meaningful precision and extends the plots to unnecessary large vertical ranges).
- Including the absolute total values in the `multi_time_pie_grid()` tables to contextualize the proportional nuclide data in the rest of the table and in the grid of pie charts.

### DIFF
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -124,7 +124,7 @@ def build_color_map(cmap_name, all_nucs=[], pivs=None):
     }
 
     if 'Other' in set(all_nucs):
-        color_map['Other'] = (0.3, 0.3, 0.3, 1.0)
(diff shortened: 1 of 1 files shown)


gonuke (approved):
Don't have a lot of opinions here - you're the one using the plots to have the best sense of the best formatting.

Some of the individual functions may have some slow scope creep, but we can review that at a different opportunity.

PR #248, opened 2026-02-25
More small updates to alaraplot

eitan-weinstein:
This PR includes a couple more small changes to `tools/alara_output_processing/alara_output_plotting.py` to improve versatility:
1) Allowing plotting of the ratios of runs against each other, by creating a new optional parameter `control_run`, wherein the user can select one of the runs contained in the ADF against which to divide to calculate a time-series of ratios.
2) Allowing for the legend to be created as a separate `matplotlib.legend.Legend` object, which can be helpful for plotting with legends with many series.
3) Allowing plotting of vertical lines at the half-life times for each nuclide plotted (can be useful because it seems that it is at the first half life where ALARA and FISPACT-II decay behaviors diverge).


gonuke (changes requested):
Looks pretty good... just few little cleanup items - thanks for the quick turnaround

gonuke (changes requested):
Just a couple of places to unwind some of the clever use of the comprehension

gonuke (changes requested):
Deleting left over commented line

PR #258, opened 2026-04-01
Addition to MT warning system to avoid warning for MTs that are automatically filtered

eitan-weinstein:
Re-opening the PR for this branch because I noticed that the ENDF-6 MT reference warning was always flagging MTs 1,2,3, and 5, which are automatically filtered out in `tendl_processing`, so it is not usefully informative to be including a warning for this case.

### DIFF
--- a/tools/ALARAJOYWrapper/preprocess_fendl3.py
+++ b/tools/ALARAJOYWrapper/preprocess_fendl3.py
@@ -9,18 +9,6 @@ from pathlib import Path
 from collections import defaultdict
 
 ISOMERIC_STATES = 'mnopqrstuvwxyz'
-PREFILTERED_MTS = {
-    1,  # (n, total)
-    2,  # (z,z0)
-    3,  # (z, nonelas.)
-    5,  # (z, anything)
-    18, # (z, fission)
-    19, # (n,f)
-    20, # (n,nf)
-    21, # (n,2nf)
-    38  # (n,3nf)
-}
-
 
 def args():
     parser = argparse.ArgumentParser()
@@ -418,7 +406,7 @@ def main():
         TAPE20.write_bytes(endf_path.read_bytes())
 
         material_id, MTs = tp.extract_endf_specs(TAPE20)
-        endf6_MTs = set(mt_dict.keys())
+        endf6_MTs = set(mt_dict)
         if len((MTs - rxd.SPEC_MTS) - endf6_MTs) > 0:
             invalid_MTs = sorted((MTs - rxd.SPEC_MTS) - endf6_MTs)
             warnings.warn(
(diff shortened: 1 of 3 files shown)


tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -443,14 +443,15 @@ def main():
         TAPE20.write_bytes(endf_path.read_bytes())
 
         material_id, MTs = tp.extract_endf_specs(TAPE20)
-        endf6_MTs = set(mt_dict.keys())
+        pre_filtered_MTs = {1,2,3,5}
+        endf6_MTs = set(mt_dict.keys()) - pre_filtered_MTs
```
gonuke: Should we just filter these out of `mt_dict` early on, e.g. when we load it from CSV in `main`? Does anything in our workflow ever need those MT numbers?

gonuke (changes requested):
Just a design question/suggestion..

tools/ALARAJOYWrapper/preprocess_fendl3.py
gonuke: I guess my thought here was to change things in `reaction_data` so that as soon as we read the MT data from the CSV, we remove the entries we don't care about.

Alternatively, we could do that right after we call the method from `reaction_data`...

I don't think we ever need these entries to be in the MT dictionary for this workflow.

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -9,6 +9,18 @@
 from collections import defaultdict
 
 ISOMERIC_STATES = 'mnopqrstuvwxyz'
+PREFILTERED_MTS = {
```
gonuke: Do you need these here still?

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -209,31 +221,36 @@ def process_mt_data(mt_dict):
...
-        if 'X' in emitted_particles:
-            emitted_particles = 'x'
-        
-        if change_NP is not None:
-            change_N, change_P = change_NP
-            data['delKZA'] = (change_P * 1000 + change_P + change_N) * 10 + M
-            data['gas'] = gas
-            data['emitted'] = emitted_particles
-        else:
+        if MT in SPEC_MTS:
```
gonuke: It might be cleaner to just do this loop twice: 
- once to delete all the ones we don't care about for any reason
- once to process what's left

gonuke (changes requested):
Quick question about deleting the MTs you don't care about

tools/ALARAJOYWrapper/preprocess_fendl3.py
eitan-weinstein: No, I just forgot to remove those

gonuke (approved):

PR #261, opened 2026-04-06
Correctly identify the top schedule

anu1217:
The changes introduced in #245 inadvertently changed the way ALARA identifies the top schedule. Before that PR, the `while` loop would terminate when a single top schedule was identified, but no error was produced when multiple top schedules were part of the input. After #245 was merged, however, the loop would no longer terminate upon finding the schedule, so the very last schedule in the list is now incorrectly considered as top. This PR attempts to fix the issue using two separate `while` loops: one to break the loop after finding the top schedule, and another to produce the error message.

### DIFF
--- a/src/Schedule.C
+++ b/src/Schedule.C
@@ -184,7 +184,6 @@ void Schedule::write(int level, char *histName, double delay, char dUnits)
   /* on first call, search for top level schedule */
   if (level==0)
     {
-      // bool found_top_sched = FALSE;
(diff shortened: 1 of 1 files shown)


gonuke:
I think there may be a simpler fix to this.  If you treat the condition of `top_sched == NULL` as the same as `!found_top_sched` you hardly have to change anything.

anu1217:
Closing due to #263 being merged.

PR #263, opened 2026-04-09
Find and report the correct top_schedule

gonuke:
This is my suggestion for an alternative to #261, which has a lighter touch by just replacing a boolean with a ptr to the schedule we are trying to find.

I also correct some paths in the sample files and update the reference output to match this new change.


src/Schedule.C
```diff
@@ -180,37 +180,39 @@ void Schedule::write(int level, char *histName, double delay, char dUnits)
...
       for (lvlNum=0;lvlNum<level;lvlNum++)
-	cout << "\t";
+	      cout << "\t";
       cout << "schedule " << ptr->schedName << " pulse_history " << histName 
-	   << " delay " << delay << " " << dUnits << " " << endl; 
+	         << " delay " << delay << " " << dUnits << " " << endl; 
     }
-
-  ptr->itemListHead->write(level+1);
+  if (top_sched != NULL)
```
anu1217, line 212: Isn't `top_sched` necessarily `!= NULL` at this point?

anu1217 (commented):
Looks good to me. I was expecting the samples to all have the correct top schedules since I generated the outputs using the `find_right_top` branch, but I believe the `bld/src/alara` executable didn't rebuild correctly in my container (figured out why).

src/Schedule.C
gonuke: Not if `level != 0`

gonuke:
Feel free to give a review & merge

anu1217 (approved):
Everything looks good to me.

PR #264, opened 2026-04-17
Moving pypact import to live within FISPACT-II class 

eitan-weinstein:
This PR follows design choices I made in #257 for external software specific Python modules (i.e. `openmc`, `openmc.deplete`, `openmc.data`) to be imported within the relevant class for the special handling for OpenMC. Likewise, I figured it appropriate to now do the same for FISPACT-II data parsing, which relies on the Python module `pypact`, whose import is now done at the start of `FispactParser.fispact_to_adf()`.


gonuke (approved):
Thanks @eitan-weinstein

PR #265, opened 2026-04-22
Allowing for Zeroing of Decay Responses after Many Half-Lives

### LINKED ISSUE #260: Find compatible decay library between ALARA and FISPACT-II (and OpenMC ?)
Over the course of the comparisons being done between ALARA and FISPACT-II for FENDL3 validation efforts, one underlying issue has been the mismatching of reference decay libraries. Currently, `ALARAJOY` inherits its decay methods from `EAFLib`, necessitating an EAF decay library to be provided. While in principle, FISPACT-II should be able to use EAF decay data as well, the two distributed decay libraries on FISPACT-II's nuclear data GitLab (https://git.oecd-nea.org/fispact/nuclear_data) are UKDD, which are ENDF-formatted. To resolve this, one of two routes could resolve this issue:

1. Locate a distribution of EAF-2010 or EAF-2012 that can be used by FISPACT-II, and have both ALARA and FISPACT-II reference EAF decay data.
2. Restructure the decay inheritance for ALARAJOY to allow for ENDF-6 formatted decay data, which should be able to process the UKDD library distributed by FISPACT-II.

If Option 1 is available,  this would likely be the preferable option because it would not require any code overhaul, although, it is not clear whether or not such a library distribution is presently available. Otherwise, even if we do include ENDF-6 decay inheritance, it should probably be done as an equal alternative option to the EAF inheritance, so that a user working with ALARAJOY could either provide an ENDF-6 or EAF reference decay library.

eitan-weinstein:
This PR includes changes to both `alara_output_processing` modules to handle decay responses for individual radionuclides at long cooling times (relative to the length their half-lives). After many half-lives have elapsed (~10) the actual values of any given decay response for a radionuclide will be effectively zero, but will continue to decay exponentially ad infinitum. Due to limitations of machine precision, though, at a certain threshold, it may be more appropriate to simply consider these values to be 0, rather than their calculated values. To resolve this, I have added a new method to the `ALARADFrame` class in `alara_output_processing/alara_output_processing.py` to set the response values for each radionuclide present in an ALARADFrame to zero after a set number of half-lives has elapsed. Additionally, I have implemented this within `alara_output_plotting.preprocess_data()` to effectuate this correction by default, with the option for user override should they wish to retain the data as-is.

I will apply the changes that I have made here to my FISPACT-II/ALARA comparisons to see if this appropriately resolves #260, in the sense of the functional issue being round-off error, as opposed to discrepancies in decay libraries, which are expected to be minimal.

### DIFF
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -16,6 +16,8 @@ def preprocess_data(
     sort_by_time='',
     pre_irrad=False,
     head=None,
+    decay_zeroing=True,
+    half_lives=10
 ):
     '''
     Prepare an ALARADFrame containing data from multiple runs and potentially
@@ -50,9 +52,17 @@ def preprocess_data(
         head (int or None, optional): Option by which to truncate the
             ALARADFrame to a particular number of rows.
             (Defaults to None)
+        decay_zeroing (bool, optional): Option to overwrite individual nuclide
+            responses after the number of half-lives set in the half_lives
+            parameter to zero to eliminate round-off error for long cooling
+            times (relative to half-life length).
+            (Defaults to True)
+        half_lives (int or float, optional): Option to specify the number of
+            half-lives to pass in cooling time before zeroing subsequent decay
+            responses. Only relevant if decay_zeroing is True.
+            (Defaults to 10)
         
     Returns:
-        times (list of floats): Chronological list of converted cooling times.
         filtered (alara_output_processing.ALARADFrame): Modified copy of input
             adf containing only rows that match all conditions in run_lbl,
             variable, and (if present) nuclides.
@@ -71,6 +81,9 @@ def preprocess_data(
         filter_dict['nuclide'] = nuclides
     
     filtered = adf.filter_rows(filter_dict)
+
+    if decay_zeroing:
+        filtered = filtered.zero_long_decay_responses(half_lives=half_lives)
 
     preset_time_unit = filtered['time_unit'].unique()[0]
     if time_unit != preset_time_unit:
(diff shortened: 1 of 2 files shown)


tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -16,6 +16,8 @@ def preprocess_data(
     sort_by_time='',
     pre_irrad=False,
     head=None,
+    decay_zeroing=True,
+    half_lives=10
```
gonuke: Could we combine these to a single parameter that is either `None` or the number of 1/2 lives.... I.e. no default?

gonuke: Or conversely, we set the default to 10 & let users turn it off by making it either very large or `None`?

gonuke (changes requested):
This looks pretty good - thanks @eitan-weinstein. I have just one question to streamline the interface.

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: Yeah that seems like a reasonable adjustment

tools/alara_output_processing/alara_output_processing.py
```diff
@@ -859,6 +859,60 @@ def create_total_rows(self):
...
+                'decay responses at these times. \n Consider increasing ' \
+                'half_lives parameter.'
+            ))
+
+        zeroed_adf = self.copy()
+
+        for nuc in set(zeroed_adf['nuclide'].unique()) - {'total'}:
+            thalf = zeroed_adf.filter_rows({
+                'nuclide' : nuc
+            })['half_life'].unique()[0]
```
gonuke, line 904: There are now at least two places where you use this pattern to get the half-life of a give nuclide.  I wonder if it makes sense to add this as a method in ADF
```
    def thalf(self, nuc):
        return self.filter_rows({'nuclide' : nuc})['half_life'].unique()[0]
```

gonuke (approved):
I've made a suggestion for convenience function for the ADF class, but will make a separate issue.

PR #266, opened 2026-04-23
Resolving ALARAJOYWrapper over-filtering of certain excitation reactions

eitan-weinstein:
Following the updates in #265, I was going back to recreate the FISPACT-II/ALARA comparisons, starting with the EU-DEMO "Activation Handbook" irradiations. When doing so, I noticed that our most recent update to the excitation handling from the final, merged version #229 resurrected a particular type of excitation issue that had been briefly resolved in an earlier iteration, though without my full understanding of why.

Recalling my first presentation of the "Activation Handbook" comparisons, I had noticed that there was a significant under-production of <sup>10</sup>Be from the irradiation of carbon. Given that the primary production pathway for <sup>10</sup>Be is <sup>13</sup>C _(n,α)_ <sup>10</sup>Be, I had assumed that this fell under the umbrella of gas handling issues, of which we had quite a few factors to resolve at that point.
carbon_february

Upon returning to this question, though, when I noticed that the <sup>10</sup>Be issue had resurfaced, I looked closer at the actual MTs contained in the <sup>13</sup>C TENDL-2017 file and saw that the relevant _(n,α)_ reaction was only logged under MT=800 (α emission with a daughter explicitly set to the ground state), without any inclusion of the cumulative MT = 107 (cumulative _(n,α)_ cross-sections). This was certainly surprising because, strictly sticking to the ENDF formatting would dictate that if there were to be any reactions in the MT 800-850 range, then MT = 107 should _necessarily_ exist. Why it does not is beyond me, but probably fits in with some of the general frustration that I have encountered since working on this project within the nuclear data community towards TENDL's inconsistency around isomer handling...

Having identified this quirk in TENDL (which plausibly exists outside of <sup>13</sup>C, but even if it is unique, is still significant, given carbon's prominence in fusion materials -- and the desire to actually process the data correctly), this PR updates the way in which ALARAJOYWrapper handles explicit excitation MTs (i.e. 800) that do _not_ have corresponding cumulative MTs (i.e. 107) to be included and have the daughter excitation levels logged as a direct function of the determinism from the MT itself.

As a quick sanity check on the legitimacy of these changes, below are plots of the <sup>10</sup>Be response for the irradiation of carbon compared to FISPACT-II, which now demonstrates very strong agreement:

be10

be10_ratios

### DIFF
--- a/tools/ALARAJOYWrapper/tendl_processing.py
+++ b/tools/ALARAJOYWrapper/tendl_processing.py
@@ -192,7 +192,7 @@ def determine_all_excitations(endf_path, MTs, pKZA, mt_dict):
 
         # Account for cases of explicit MF3 excitation reactions without
         # corresponding cumulative MTs (i.e. (n,a0) [MT=800] for C-13)
-        elif cumulative_MT not in MTs:
+        elif cumulative_MT is not None and cumulative_MT not in MTs:
             M = np.where(EXCITATION_DICT[cumulative_MT] == MT)[0][0]
             # MT = 50 is forbidden in ENDF6, so scattering reactions require
             # an index adjustment compared to other state-explcit reactions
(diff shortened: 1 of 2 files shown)


eitan-weinstein:
Log of other cases that this PR has resolved (I'll continue to update as I make my way through these irradiation comparisons):

- <sup>58m</sup>Co production in Ni irradiation (t_1/2 = 8.85 h); _"Activation Handbook"_
  -  Moderate contributor to specific activity at shutdown, shortly thereafter (~17%)
  - Essentially completely suppressed prior to update
  - Near perfect agreement (≤0.1%) between FISPACT-II/ALARA now
- <sup>178n</sup>Hf production in W irradiation (t_1/2 = 31 y); _"Activation Handbook"_

tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -374,51 +391,61 @@ def iterate_MTs(
...
-
-                if special_MT not in all_rxns[pKZA][dKZA]:
-                    all_rxns[pKZA][dKZA][special_MT] = {
-                        'emitted'  : emitted,
-                        'xsections': np.zeros(VITAMIN_J_ENERGY_GROUPS)
+    for MT in MTs:
+        if (
+            MT not in EXCITATION_REACTIONS
+            or REVERSE_EXCITATION_DICT[MT] not in MTs
+        ):
```
gonuke: I think I'd like to separate this into two loops. 
1. loop to filter the list of `MTs` down to th eones that we'll actually process
2. loop over those filtered MTs

This will make it easier to grapple with this filtering.
```
    filtered_MTs = {}
    for MT in (MTs - EXCITATION REACTIONS):
        if REVERSE_EXCITATION_DICT[MT] not in MTs
            filtered_MTs.add(MT)

    for MT in filtered_MTs:
```

gonuke (changes requested):
Clarifying logic vs loop efficiency.

gonuke (approved):

PR #267, opened 2026-04-24
Small alaraplot updates -- revisited after rebase

eitan-weinstein:
This PR takes over for #248. Upon attempting to rebase to be able to respond to an old change request on there from last month, I ran into some trouble with the Git history. Given that the most recent requested change was just to remove a single comment line, I figured it was not worth going through and untangling that Git mess, instead opting just to copy over the latest version of `alara_output_plotting.py` from the old PR to here and removing said line.

Below is the original description for #248:

> This PR includes a couple more small changes to tools/alara_output_processing/alara_output_plotting.py to improve versatility:

>1. Allowing plotting of the ratios of runs against each other, by creating a new optional parameter control_run, wherein the user can select one of the runs contained in the ADF against which to divide to calculate a time-series of ratios.

>2. Allowing for the legend to be created as a separate matplotlib.legend.Legend object, which can be helpful for plotting with legends with many series.

>3. Allowing plotting of vertical lines at the half-life times for each nuclide plotted (can be useful because it seems that it is at the first half life where ALARA and FISPACT-II decay behaviors diverge).

### DIFF
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -2,6 +2,8 @@ import numpy as np
 import pandas as pd
 import matplotlib.pyplot as plt
 from matplotlib import lines
+import matplotlib.cm as cm
+import matplotlib.colors as mcolors
 from warnings import warn
 import alara_output_processing as aop
 
@@ -106,7 +108,7 @@ def preprocess_data(
 
     return filtered, piv
 
-def build_color_map(cmap_name, all_nucs=[], pivs=None):
+def build_color_map(cmap_name, all_nucs=[], pivs=None, mark_thalf=False):
     '''
     Given a list of pivot DataFrames (one per run) or a 1D array-like data
         structure of nuclide string names, build a stable color mape keyed by
@@ -133,9 +135,14 @@ def build_color_map(cmap_name, all_nucs=[], pivs=None):
     if len(all_nucs) == 0:
         raise ValueError('Must input either all_nucs or pivs.')
 
-    color_map = {
-        lbl: cmap(i % cmap.N) for i, lbl in enumerate(sorted(all_nucs))
-    }
+    if mark_thalf:
+        color_map = {
+            lbl: cmap(0.4 + 0.55 * i / max(len(all_nucs)-1,1)) for i, lbl in enumerate(sorted(all_nucs))
+        }
+    else:
+        color_map = {
+            lbl: cmap(i % cmap.N) for i, lbl in enumerate(sorted(all_nucs))
+        }
 
     if 'Other' in set(all_nucs):
         color_map['Other'] = (0.8, 0.8, 0.8, 1.0)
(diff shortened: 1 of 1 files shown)


tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -92,7 +94,7 @@ def preprocess_data(
 
     return filtered, piv
 
-def build_color_map(cmap_name, all_nucs=[], pivs=None):
+def build_color_map(cmap_name, all_nucs=[], pivs=None, mark_thalf=False):
```
gonuke, line 110: Need additional docstring for new parameter

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -557,34 +598,73 @@ def plot_single_response(
...
+                # Conditional vectorized division to calculate time-series
+                # ratio against the control run. If zeros exist in the control
+                # run, a zero-division RuntimeWarning will be raised, but does
+                # not cause plotting issues as NaNs will just not be plotted.
+                # If a ratio series starts/stops abruptly, this zero-division
+                # is the cause and is not necessarily an error, as various
+                # nuclides may not be present across all cooling times.
+                y = (
+                    piv.loc[nuc].to_numpy() / control_piv.loc[nuc].to_numpy()
+                    if ratio_plotting else piv.loc[nuc].tolist()
```
gonuke: Is there a specific reason that this is a list and not a numpy array?
```suggestion
                    if ratio_plotting else piv.loc[nuc].to_numpy()
```

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -557,34 +598,73 @@ def plot_single_response(
...
+                )
+            except KeyError:
```
gonuke: This doesn't seem to catch the `RuntimeWarning` that you mentioned.  Should it?

It seems more like the purpose of this `try... except...` block is for the `KeyError`.  Can't we tests for that `KeyError` explicitly rather than relying on the exception handling?

Something like (pseudo-code):
```
if nuc not in run_lbl:
    print("Warning")
    continue

y = piv.loc[nuc].to_numpy()

if ratio_plotting:
    # maybe this needs its own `try...except...` block???
    y /= control_pic.loc[nuc].to_numpy()
```

gonuke (changes requested):
Just trying to clean up the readability of some of the logic

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -592,34 +637,73 @@ def plot_single_response(
...
                 label=(nuc + label_suffix),
                 color=color_map[nuc],
                 style=style
             )
 
-    title_suffix = f'{variable} vs Cooling Time '
+            if mark_thalf and nuc not in plotted_nucs:
+                thalf = filtered.loc[
+                    filtered['nuclide'] == nuc, 'half_life'
+                ].iat[0]
```
gonuke: Does this need an update or rebase following #269?

gonuke (changes requested):
One update for the new method to get half lives.

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: Good catch -- I'll rebase now

tools/alara_output_processing/alara_output_plotting.py
gonuke: Should this be `run_lbls` now and indicate that it can be more than one?

gonuke (changes requested):
One docstring fix...

gonuke (approved):

Issue #204, opened 2025-12-16
Handle ALARAJOY Isomers Missing Decay Data


eitan-weinstein:
A number of key nuclides in FENDL-3.2x (TENDL-2017) contain cross-section data for (n,nx) reactions (MT 50-91), leaving behind a residual nucleus whose energy level ranges from a ground state up to a 40<sup>th</sup> excited metastable state. While some of these cross-sections correspond to daughter nuclides that may plausibly appear in fusion-relevant materials for some small time post-irradiation (i.e. <sup>183m</sup>W and <sup>16m</sup> N from EUROFER-97), many are either too highly excited or short-lived to have any decay data on them. As discussed in a comment on #194, this can create a number of issues, chiefly that an isomer could be produced that should decay nearly instantaneously, but lacking proper reference decay data, appears to be a stable nuclide and does not decay at all.

Handling this physical problem of isomer decay can be done by one of the two approaches:

1. Modify the isomer saving in `ALARAJOYWrapper` to reference each individual isomer against an EAF decay library to see if that isomer exists in the library and has decay data.

      a. EAF is specifically needed because `ALARAJOY` is a hybrid-library that borrows its decay data from EAF, given that TENDL-2017 only has activation data, not decay data.

2. Manually force the decay of these isomers, which may well exist at shutdown, by rewriting their number densities for any time after shutdown to 0, compensated by adding the number density values to their respective ground state nuclides in ALARA.

     a. I am disinclined to pursue this approach both because it would necessitate changes to ALARA's source code, when in reality, this issue is caused by ALARAJOY's inability to distinguish between physically relevant and irrelevant isomers (notably other data sources create very few metastable daughters). Furthermore, I do not think that there would be a reasonably physically appropriate method to retroactively convert these number densities to their ground state nuclides without tampering with decay process already undergoing post-shutdown for those nuclides.

bohmt:
This FISPACT-II wiki page briefly mentions what they are using for decay data:
https://fispact.ukaea.uk/wiki/Nuclear_data_forms#Decay_Data
And I can confirm 3875 files in my ...fispactInstallpath/ENDFdata/decay/decay_2012/

PR #271, opened 2026-05-07
Opening capabilities for UKDD decay library processing

### LINKED ISSUE #260: Find compatible decay library between ALARA and FISPACT-II (and OpenMC ?)

eitan-weinstein:
Closes #260 (after reopening it too).
)
This PR expands ALARAJOYWrapper's ability to handle decay libraries beyond EAF distributions to UKDD (i.e. `decay_2020`) decay libraries. This follows the long-standing discussions between @gonuke, @bohmt, and myself and I finally was able to figure out how to get this workable. This will not only allow for better comparisons against FISPACT-II for our forthcoming paper, but will also allow for ALARAJOYWrapper to more closely follow the data recommendations in FENDL3.2x, which specify the use of `decay_2020` as the preferred decay library.

The key issues with `decay_2020` that prevent it from being immediately applicable to ALARAJOY are that it is distributed in files for individual nuclides which need to be compiled (minor) and that many of these files have distinct formatting issues (more major) that prevented them from being readable by `EAFLib.C`, which should be able to read any properly formatted ENDF/B file. I made use of the NEA tool `endf-parserpy` and it's custom errors to diagnose the issues with any individual UKDD file, which are patched and then compiled to a single decay library file that is readable by ALARAJOYWrapper and `EAFLib.C` with only minor changes to the latter.


eitan-weinstein:
I've moved this PR back to being a draft because I ran into an issue of certain radionuclides being reported as stable in inventories calculated with ALARA binary libraries produced from `decay_2020` with these processing methods. Upon further investigation, I found that this issue was coming from two sources:

1. Ground-state radionuclide products of reactions contained in the TENDL data that are _not_ present in the decay library.
2. Nuclides (seemingly only isomers, at least so far) whose half-lives are being correctly parsed by `EAFLib.C`, but for some reason have malformed data that is leading them to ultimately be written out to the binary library as being stable.


For the first issue, this relates back to the conversations from #204 , but our course of action was predicated on the assumption that we would only have this kind of mismatch for isomeric daughters. Importantly, this issue is not new to the changes introduced here, but I only noticed it now because the number densities of these exotic radionuclides were generally small enough to not make meaningful contributions in the aggregate results and slipped under the radar in the various analyses I've been conducting. However, for the "Activation Handbook" irradiation of tungsten, which is the case that first alerted me to these issues now, there were 4 radionuclides present in the inventory with reported half-lives of `-1` (stable) that were simply missing both from the previously used EAF decay library and the current UKDD decay library (<sup>180</sup>Tm, <sup>182</sup>Yb, <sup>183</sup>Yb, and <sup>186</sup>Lu). While their number densities were relatively small, their inclusion as falsely stable nuclides (surely amongst others as well, though I have not yet compiled a complete list of all of these cases), is obviously untenable and nonphysical. It seems to me that the most physically appropriate solution would have to be discarding the reactions that produce these products entirely in `tendl_processing.iterate_MTs()` because, unlike the isomer deexcitation method that we apply for the similar case of isomers lacking decay data, we do not have any physically reasonable basis for an assumed near-immediate decay pathway to a determinate nuclide/state. While it would be preferrable not to have to discard any data from the TENDL cross-sections, we may simply be limited by the state of the nuclear data sources themselves. I would certainly be open to further discussion on this matter @gonuke and @bohmt to see if we should handle this type of edge case differently, but for now I'll implement a discarding method for such nuclides to this PR.

For the second issue, I am still working to identify where under-the-hood in ALARA these other half-lives are being either lost or overwritten. I am confident that this is not an issue with ALARA's source code but rather with the UKDD processing introduced in this PR because these nuclides (<sup>177m</sup>Lu, <sup>177m</sup>Hf, <sup>177n</sup>Hf, <sup>178n</sup>Hf, <sup>179n</sup>Hf, <sup>182n</sup>Ta, and <sup>191n</sup>Ir for the W "Activation Handbook" irradiation) all have the correct half-lives in the old ALARAJOY ALARA binary library that used EAF decay data. I know that `EAFLib.C` necessitates ENDF/B formatting and my use of `endf-parserpy` should be ensuring compliance to ENDF formatting standards, but there must be some other factor that this does not cover, so I will continue digging in to resolve this.

eitan-weinstein:
My suspicion that this second false-stability error source was only present for isomers appears to be correct, with the following being the total list of UKDD-2020 isomers causing this issue, many of whom appeared in the tungsten results:

1. <sup>53m</sup>Fe
2. <sup>98m</sup>Y
3. <sup>93m</sup>Mo
4. <sup>152n</sup>Eu
5. <sup>154m</sup>Eu
6. <sup>177m</sup>Lu
7. <sup>177m</sup>Hf
8. <sup>177n</sup>Hf
9. <sup>178n</sup>Hf
10. <sup>179n</sup>Hf
11. <sup>190m</sup>Os
12. <sup>191n</sup>Ir
13. <sup>192n</sup>Ir
14. <sup>185m</sup>Pt
15. <sup>196n</sup>Au
16. <sup>198m</sup>Au
17. <sup>206m</sup>Tl
18. <sup>202m</sup>Pb
19. <sup>204m</sup>Pb
20. <sup>212m</sup>Po

Tellingly, I was able to identify these nuclides as the culprits as they all have two entries in the ALARA library index, one containing all of the activation reaction pathways and one with the decay pathways, when in reality they should be singular. Between these, the latter listed the correct half-lives, while the former labeled them as stable. Knowing this, I should be able to identify -- and hopefully resolve relatively painlessly -- the source of this duplication within the original UKDD files where their decay data is contained.

eitan-weinstein:
Interestingly, for all of the above-listed isomers, when `convert_lib` is run with their _individual_ files from the uncompiled `decay_2020/` directory, the ALARA binary index appropriately contains a single section with the correct half-life for the given nuclide, meaning the issue is arising from the compilation itself, rather than the in-situ file reformatting that occurs ahead of the appending step.

eitan-weinstein:
Ok it looks like I've resolved this isomer issue! In my previous processing before my most recent commit ([`041d3b5`](https://github.com/svalinn/ALARA/pull/271/commits/041d3b5bbb19193edd9b8706910b405218ca7ae9)), I had excluded the `TEND` record from each individual file (except the actual final file) while being appended to the compiled file. My reasoning was in trying to keep with proper ENDF formatting, given that we are making a single large "tape". Nevertheless, after toying with its removal, the transmutation/decay duplication issue disappeared and `EAFLib.C` looks to be able to appropriately handle this decay data now. 

While this outcome is not entirely satisfying because it does not explain _why_ this issue was only popping up for these 20 isomers and no other nuclides, given that the `TEND` removal was applied uniformly.

Nevertheless, I'm re-opening this PR now and will move forward with the reproduction of my analysis plots for the FNS and "Activation Handbook" studies.

eitan-weinstein:
As further confirmation of the correctness of these updates, here's the new comparison of nuclide contributions to decay heat for the tungsten irradiation that initially drew my attention to the issues guiding my work for the past few days:

image


Aside from minute differences in contribution percentages, the responses for ALARA and FISPACT-II for this tungsten irradiation are now nearly identical and the hafnium isomers are behaving appropriately.

src/DataLib/EAFLib.C
```diff
@@ -397,14 +397,28 @@ void EAFLib::getDecayInfo()
...
+    {
+      lineStart = inDecay.tellg();
+      if (!inDecay.getline(buffer, MAXLINELENGTH)) break;
+
+      int len = (int)strlen(buffer);
+      if (len < 75) continue;
+
+      char mtField[4] = {0};
+      strncpy(mtField, buffer + 72, 3);
+      MT = atoi(mtField);
```
gonuke: Can we write a little function that does this and returns -1 if `len < 75`?

```
int parse_MT(const char* buffer) {
    int MT_start = 72;
    int MT_length = 3;

    if (strlen(buffer) < MT_start + MT_length) return -1;

    char mtField[4] = {0};
    strncpy(mtField, buffer+MT_start, MT_length);
    return atoi(mtField);
}
```

Then we can get the first value of MT before the while loop and make it `while (MT != 457)`

src/DataLib/EAFLib.C
```diff
@@ -438,8 +452,11 @@ int EAFLib::getDecayData()
       /* scan for first 457 card */
       inDecay.getline(buffer,MAXLINELENGTH);
       /* extract the MT number of this line */
-      buffer[75] = '\0';
-      MT = atoi(buffer+72);
+      int len = (int)strlen(buffer);
+      if (len < 75) { MT = 0; continue; }
+      char mtField[4] = {0};
+      strncpy(mtField, buffer + 72, 3);
+      MT = atoi(mtField);
```
gonuke: See above about a new private function for this

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -293,63 +470,74 @@ def get_MT_from_line(line):
...
-            _in_decay_block = False
-            line = f.readline()
+        _in_decay_block = False
+        line = f.readline()
 
-            while line:
+        while line:
 
-                # Radionuclide processing
+            try:
```
gonuke: Is there a way to test this without a try block? From the comment below, we know all the ways this may fail and could write a function that tests for those explicitly?

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -252,13 +260,182 @@ def process_mt_data(mt_dict):
...
+    with open(decay_file, 'r') as f:
+        lines = f.readlines()
+
+    with open(compiled_file, 'a') as f:
+        f.writelines(lines)
+
+def ensure_line_length_compliance(compiled_file):
+    """
+    Split lines that may have been joined together without a newline separator
+        during file compilation.
```
gonuke: Can't we just avoid introducing this in the first place?

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -252,13 +260,182 @@ def process_mt_data(mt_dict):
 
     return mt_dict
 
-def eaf_float(num_str):
+def resolve_ukdd_inconsistencies(decay_parser, decay_file):
```
gonuke: Can't we test for these things as we do the concatenation?  

* We read all the lines in - we could look for blank lines and delete them before appending them to the compiled file.
* We can detect whether the TEND_RECORD is missing and add it
* We can skip the binary file based on it's name or some other characteristic

gonuke (changes requested):
Sorry for the delay in getting to this.  A few questions to make this a little cleaner, I hope.

src/DataLib/EAFLib.C
```diff
@@ -393,18 +393,45 @@ int EAFLib::getTransData()
...
+    if (!inDecay.getline(buffer, MAXLINELENGTH))
+      break;
+
+    MT = parse_MT(buffer);
+  }
+
+  if (MT == 457)
+  {
+    inDecay.seekg(lineStart);
+  }
```
gonuke, line 434: De we need to do aynthing if `MT != 457`?  I'm guessing that's only possible if we've reached the end of the file?  Is that just taken care of elsewhere?

tools/ALARAJOYWrapper/reaction_data.py
eitan-weinstein: I was able to avoid relying on `try`/`except` for the UKDD formatting fixes, but `EAFLib` is expecting the nuclides to be listed in the file by ascending KZA, so I think the separate loops for reformatting/KZA extraction and then concatenation may still be needed.

Alternatively, we could pre-determine the KZA names from the filenames (which should all be appropriately named nuclides), but it makes us dependent on the file metadata being correct, rather than relying directly on the data itself.

gonuke (commented):
Found another question about the logic/flow.

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -252,13 +260,113 @@ def process_mt_data(mt_dict):
...
+            lines = f.readlines()
+
+        # Remove blank lines
+        updated_lines = [line for line in lines if line.strip()]
+
+        # Include missing TEND Records
+        if TEND_RECORD.strip().rstrip('0').strip() not in updated_lines[-1]:
+            tend_line = TEND_RECORD
+            if not updated_lines[-1].endswith('\n'):
+                tend_line = '\n' + tend_line
```
gonuke: Should this have a trailing `'\n'`?

tools/ALARAJOYWrapper/reaction_data.py
gonuke: I think the logic would be cleaner if we had one function that made a map of kza -> filename and then another function that compiles all the files.

We should make that map regardless of the source of the data.  In some cases it may be trivial, but it makes for more consistent data structures and fewer complicated branches.

src/DataLib/EAFLib.C
eitan-weinstein: No, we shouldn't need any extra handling for `MT != 457` because the format of each nuclide's decay data within a compiled file will always be:

```
MF1, MT451
MF8, MT457
EOF
```
So there should not be any case where we run into more data for a given nuclide after we no longer have `MT == 457`.

eitan-weinstein:
Now that I'm no longer relying on `endf-parserpy` for its custom exceptions, my most recent commit detaches these updates from being dependent on it, instead using the `ENDFtk`-based `create_endf_file_obj()` function from `tendl_processing.py`. In order to avoid circular import errors, `tendl_processing.create_endf_file_obj()` is imported within `reaction_data.compile_decaylib()`.

On top of minimizing dependencies, it seems that extracting the KZA with `ENDFtk` is much faster than `endf-parserpy`, which is a nice added benefit.

gonuke (changes requested):
A couple of additional comments while you're in the flow...

tools/ALARAJOYWrapper/reaction_data.py
eitan-weinstein: In my current handling, the trailing `'\n'`  is checked and conditionally added in `append_to_compiled_lib()` because we may have a case in which the `TEND_RECORD` is present in the file, but only missing the trailing `'\n'`, which is distinct from the cases where the `TEND_RECORD` is missing entirely. Either way, the final line from an individual nuclide file will be appended to the compiled file with a trailing `'\n'`, but my current approach separates a missing EOF from a missing final newline.

tools/ALARAJOYWrapper/reaction_data.py
eitan-weinstein: In principle, I think that is reasonable, but the EAF distributions (at least 4.1 and 2010 which are the only ones I can find online) are not organized by individual nuclide, but rather are split into 10 compiled decay data files, which are already in order of ascending KZA. It may still make sense for separation of concerns to make this KZA/filename map separately, but it would only apply to the UKDD data. Additionally, in doing so, we would still have to apply blank line and TEND record fixes first to get the ENDF formatting correct before parsing data with `ENDFtk` should we choose to calculate KZA from the file itself rather than its name. Otherwise, we could just get the KZA information from the filename like we already do for the cross-section data and apply all formatting fixes in the compilation loop.

I do think that moving away from specific filename formatting is probably for the best anyways, and this endeavor is leading me to believe it to be worthwhile also to implement a similar approach for cross-section data files too because it is less fragile being dependent only on the actual data themselves.

tools/ALARAJOYWrapper/reaction_data.py
gonuke: I get that this is checked elsewhere, but given that you consider the missing `'\n'` to be pathological, it seems to make sense to add it here when you know you need it rather than let your later test put it in where it's missing.

tools/ALARAJOYWrapper/README.md
```diff
@@ -22,20 +22,39 @@ This preprocessor uses [NJOY 2016](https://github.com/njoy/NJOY2016) Nuclear Dat
   * [Pandas](https://pandas.pydata.org/docs/getting_started/install.html)
 - Domain-specific packages
   * [ENDFtk](https://github.com/njoy/ENDFtk)
-  * [NJOY 2016](https://github.com/njoy/NJOY2016) (built with `develop` branch)
+  * [NJOY 2016](https://github.com/njoy/NJOY2016) (at Git hash `1dbce787`)
```
gonuke: Can you put a footnote somewhere about why this specific hash?  Presumably, some new release will incorporate this and the specific hash won't be necessary?

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -293,63 +394,62 @@ def get_MT_from_line(line):
...
+        line = f.readline()
+
+        while line:
+            MT = get_MT_from_line(line)
+            if not _in_decay_block and MT == DECAY_MT:
+                _in_decay_block = True
+
+                # Parse nuclide KZA
+                za = int(standardize_float(line[:11]))
+                M = int(line[33:44].strip())
```
gonuke: `strip()` seems unnecessary here if you are selecting a particular substring?

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -293,63 +394,62 @@ def get_MT_from_line(line):
...
+                kza = za * 10 + M
+                stability = int(line[44:56].strip()) # 0->unstable , 1->stable
```
gonuke: `strip()` seems unnecessary here if you are selecting a particular substring?

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -252,13 +255,111 @@ def process_mt_data(mt_dict):
...
+    from tendl_processing import calculate_KZA_from_ENDF
+
+    ukdd_options = ['ukdd', 'ukaeadd', 'decay_2020', 'decay_2012']
+    if decay_lib_type.lower() in ukdd_options:
+        decay_lib_type = 'ukdd'
+
+    for decay_file in decay_dir.iterdir():
+        if '.DS_Store' in str(decay_file):
+            decay_file.unlink()
+            continue
```
gonuke: This should be resolved in some other way.  Can we match positively on some pattern and just ignore everything else in case other files end up in there?

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -252,13 +255,111 @@ def process_mt_data(mt_dict):
...
+        # Include missing TEND Records
+        if TEND_RECORD.strip().rstrip('0').strip() not in updated_lines[-1]:
+            tend_line = TEND_RECORD
+            if not updated_lines[-1].endswith('\n'):
+                tend_line = '\n' + tend_line
+            updated_lines.append(tend_line)
+
+        # Ensure file ends with a newline
+        if not updated_lines[-1].endswith('\n'):
+            updated_lines[-1] += '\n'
```
gonuke: Changing the order of these makes them a little simpler:
```suggestion
        # Ensure file ends with a newline
        if not updated_lines[-1].endswith('\n'):
            updated_lines[-1] += '\n'
            
        # Include missing TEND Records
        if TEND_RECORD.strip().rstrip('0').strip() not in updated_lines[-1]:
            updated_lines.append(TEND_RECORD + "\n")
```

Could be even simpler if you add the `\n` to `TEND_RECORD` above

gonuke (changes requested):
Some final cleanup

tools/ALARAJOYWrapper/README.md
eitan-weinstein: I just went back to double check NJOY and since my communication with the NJOY team where they pointed me to this update, they have indeed merged this update. I'll just remove this hash specification.

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -252,13 +255,112 @@ def process_mt_data(mt_dict):
...
+            # Remove blank lines
+            updated_lines = [line for line in lines if line.strip()]
+
+            # Ensure file ends with a newline
+            if not updated_lines[-1].endswith('\n'):
+                updated_lines[-1] += '\n'
+
+            # Include missing TEND record
+            if (
+                TEND_RECORD.strip().rstrip('0').strip()
```
gonuke: Maybe this should be a regex again?  Or leave off the newline after all in `TEND_RECORD` and add it when you append `TEND_RECORD`?

gonuke (changes requested):
seems like a lot of `strip` gymnastics now for `TEND_RECORD` matching...

tools/ALARAJOYWrapper/reaction_data.py
```diff
@@ -252,13 +255,109 @@ def process_mt_data(mt_dict):
...
+
+            # Remove blank lines
+            updated_lines = [line for line in lines if line.strip()]
+
+            # Ensure file ends with a newline
+            if not updated_lines[-1].endswith('\n'):
+                updated_lines[-1] += '\n'
+
+            # Include missing TEND record
+            if TEND_RECORD.rstrip('0').strip() not in updated_lines[-1]:
```
gonuke, line 300: does this just evaluate to the string `'-1 0  0'` now?  If so, is that specific enough?  Could a file have that string and not be a correctly formatted last line?

gonuke (changes requested):
Sorry - a little more clarity needed  on the TEND_RECORD thing...

tools/ALARAJOYWrapper/reaction_data.py
eitan-weinstein: That's correct that the string operations evaluate to `'-1 0  0'`, but it should not be possible that an ENDF-formatted file would contain this string in any other context other than at the end of the "tape". Canonically, the actual TEND record is defined only as this string, but for formatting purposes I've included the whitespace in the constant definition. Nevertheless, the sequence `'-1 0  0'` can only be found once per legitimate ENDF file.

gonuke (approved):
Should all be good now... thanks @eitan-weinstein

gonuke:
Apparently there are merge conflicts...

eitan-weinstein:
Just rebased with all of the newly merged updates, so I think it should be good to go now.

gonuke (approved):

PR #272, opened 2026-05-11
Fix syntax errors in sched_post_processor.py

anu1217:
There were some syntax errors and redundant function arguments that seem to have slipped through the cracks when the sched_post_processor.py tool was added. The processor also did not correctly account for extra lines in the output that come from added verbosity, due to the `while not line.startswith("pulse_history")` condition being overly generic.

### DIFF
--- a/tools/sched_post_processor.py
+++ b/tools/sched_post_processor.py
@@ -59,7 +59,7 @@ def make_sch_sub_dict(sch_line):
     return sch_sub_dict
 
 
-def make_pe_sub_dict(pe_line, unit_multipliers):
+def make_pe_sub_dict(pe_line):
     pe_sub_dict = {
         "type" : "pulse_entry",
         "pe_dur": float(pe_line[1]) * unit_multipliers[pe_line[2]],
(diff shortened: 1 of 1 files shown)


tools/sched_post_processor.py
```diff
@@ -77,33 +77,29 @@ def make_nested_dict(lines):
...
+    while any(lines[line_idx].strip().startswith(keyword) for keyword in keywords):
         new_child_level = lines[line_idx].count("\t")
         tokens = lines[line_idx].strip().split()
-
-
         while new_child_level < len(ancestors):
             current_sched = ancestors.pop()
-
-        if tokens[0] == "schedule":
+        if tokens[0] == keywords[1]:
```
gonuke: This is a dangerous change if anyone chooses to reorder `keywords`.  If you want to avoid magic strings (although that's less important than magic numbers), you could define a variable for each of these keywords and then assemble the `keywords` list from those variables.

gonuke (changes requested):
Do we have tests for this?  I'm guessing that if we did, they would fail this. (Maybe we don't have any tests for any of the ALARA tools?)

anu1217:
I don't think we have tests for most of the tools in this repo, but it would definitely be useful to have them, especially as these tools are likely to be frequently used and updated

gonuke (approved):

PR #273, opened 2026-05-13
Including Mean and Standard Deviation for alaraplot ratio plotting

eitan-weinstein:
Follows #267.

This PR introduces a small change to the function `alara_output_plotting.plot_single_response()` for the time-series ratio plotting functionality to include the mean and standard error of the mean in the legend for each series. This offers a quantitative view to the similarity of the sample series against the control run, which I think will both be valuable for me as I am writing the ALARAJOY updates paper, and generally valuable for users.

### DIFF
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -704,7 +704,7 @@ def plot_single_response(
             y = piv.loc[nuc].to_numpy()
             if ratio_plotting:
                 y /= control_piv.loc[nuc].to_numpy()
-                if not np.isnan(y.mean()):
+                if not np.isnan(y.mean()) and nuc == 'total':
                     mean = format_statistic(y.mean(), sig_figs)
                     sem = format_statistic(y.std() * len(y) ** -0.5, sig_figs)
                     label_suffix += (
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -530,7 +530,7 @@ def plot_single_response(
     plot_type='plot',
     separate_legend=False,
     control_run=None,
-    sig_figs=4,
+    sig_figs=3,
     mark_thalf=False
 ):
     '''
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -174,6 +174,8 @@ def split_label(label):
         parts = label.split('(')
         isotope = parts[0].strip()
         run_lbl = f'({parts[1].strip(')')})'
+        if '$\\mu' in run_lbl:
+            run_lbl = run_lbl[:-1]
     else:
         isotope = label.strip()
         run_lbl = ''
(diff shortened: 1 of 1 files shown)


tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -201,6 +203,30 @@ def reformat_isotope(isotope):
...
+
+    Returns:
+        formated_statistic (str): Value with the desired number of significant
+            figures, either in fixed-point or scientific notation
+    '''
+
+    return (
+        f'{statistic:.{sig_figs}f}'
+        if statistic > 1e-3 and statistic < 1e3
+        else f'{statistic:.{sig_figs}e}'
```
gonuke: You could try using `g` instead of `f` or `e` to handle this automatically (and maybe not need a function)

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -174,6 +174,8 @@ def split_label(label):
         parts = label.split('(')
         isotope = parts[0].strip()
         run_lbl = f'({parts[1].strip(')')})'
+        if '$\\mu' in run_lbl:
```
gonuke, line 177: What's this about?? Why are we only dropping the last character of this label?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -671,8 +704,14 @@ def plot_single_response(
             y = piv.loc[nuc].to_numpy()
             if ratio_plotting:
                 y /= control_piv.loc[nuc].to_numpy()
+                if not np.isnan(y.mean()) and nuc == 'total':
+                    mean = format_statistic(y.mean(), sig_figs)
+                    sem = format_statistic(y.std() * len(y) ** -0.5, sig_figs)
```
gonuke: Not sure where this formula comes from... why multiply by `sqrt(len(y))`? (Also may be better to use `np.sqrt()`)

gonuke (changes requested):
A few clarifying questions....

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: I was trying to implement the formula for standard error from https://en.wikipedia.org/wiki/Standard_error

$\sigma_{\mu} = \frac{\sigma}{\sqrt{n}}$

With n being the number of values in the series.

That said, if you think another statistical metric may be more appropriate, like coefficient of variation or just the unmodified standard deviation itself, I can put that in/replace it.

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: I put this in because the `run_lbl` in this case is formatted as `f"{element}-{A} ({run_lbl})"`, which would result in there being an extra `")"` included after the statistics. I can more explicitly implement this, though, by having the lines be:

```
if '$\\mu' in run_lbl:
      run_lbl = run_lbl.strip(')')
```

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: I think the standard deviation would actually probably be the best anyways. I'll simplify things.

gonuke (approved):
I think this looks good, although I've definitely lost clarity on all the string modifications

PR #274, opened 2026-05-13
Fixing p0 and d0 reaction filtering in ALARAJOYWrapper

eitan-weinstein:
Previously, in #229, I must have accidentally not included MT 600 and 650 in the `EXCITATION_DICT` used for filtering duplicate excitation reaction pathways if the cumulative reactions are present as well. I noticed this problem because I was seeing an unreasonably high amount of <sup>14</sup>C production from the simulated irradiation of nitrogen, caused by the <sup>14</sup>N _(n,p)_ <sup>14</sup>C reaction. Upon looking to my DSV file, I saw duplicates of both the `70140 60140 p` and `70140 60130 d` reactions, explaining the overproduction. As it stood, `EXCITATION_DICT` did not include the explicit (n,p0) and (n,d0) as it should have, allowing them to sneak past the filtration system in `iterate_MTs()`. After this patch, the nitrogen reactions all look ok and I'm sure there are reactions for other elements that are affected by this.

### DIFF
--- a/src/DataLib/EAFLib.C
+++ b/src/DataLib/EAFLib.C
@@ -58,7 +58,7 @@ void EAFLib::extract(char* input, float* value)
   char section[32];
 
   /* find beginning of exponent */
-  int expStart = strcspn(input,"+-");
+  int expStart = strcspn(input+1,"+-") + 1;
 
   /* extract mantissa portion to string variable */
   strncpy(section,input,expStart);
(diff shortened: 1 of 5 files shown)


gonuke:
I expect this to be a small patch, but it appears to include a lot of other code...

eitan-weinstein:
> I expect this to be a small patch, but it appears to include a lot of other code...

Oh, I think I built this branch from #271, but this should be independent of that. I'll open a new PR with the minimal fix.

PR #277, opened 2026-05-17
Fixing FISPACT-II cooling time parsing

eitan-weinstein:
This PR fixes a bug in `alara_output_processing.FispactParser.fispact_to_adf()` in which the cooling times actually corresponded to the _accumulated_ sum of all time passed at that cooling time, rather than the cooling time itself. This is due to the way that `pypact` processes these cooling times, which can be seen in[ `pypact.filerecord.FileRecord._process()`](https://github.com/fispact/pypact/blob/master/pypact/filerecord.py#L91) in line 107:

```
        self.cooling_times = list(accumulate(self.cooling_times))
```

I'm not exactly sure why `pypact` does this, but I noticed this issue when comparing the responses from ALARA and FISPACT-II, and noticing that FISPACT-II's responses extended much further than ALARA's, even though my inputs for both simulations used the exact same cooling times. After implementing the adjustment from this PR, the cooling times from both sets of data agreed as they should.

### DIFF
--- a/tools/alara_output_processing/alara_output_processing.py
+++ b/tools/alara_output_processing/alara_output_processing.py
@@ -379,6 +379,7 @@ class FispactParser:
             row = {k: -1 for k in ['block', 'block_name', 'block_num']}
             row['run_lbl'] = run_lbl
             row['time_unit'] = time_unit
+            previous_times = 0.0
             for time in output.inventory_data:
                 
                 # FISPACT-II marks both pre-irradiation and shutdown as
@@ -386,11 +387,23 @@ class FispactParser:
                 # instance (pre-irradiation) to -1 to match the ALARADFrame
                 # time label convention.
                 if time.cooling_time == 0 and time_zero:
-                    time.cooling_time = -1
-                    mass = sum(nuc.grams for nuc in time.nuclides) / 1e3 # [kg]
+                    actual_cooling_time = -1
+                    mass = sum(nuc.grams for nuc in time.nuclides) / 1e3 #[kg]
                     time_zero = False
+                
+                # The Pypact value time.cooling_time is a cumulative value of
+                # all cooling times up to the actual cooling time in
+                # reference. In order to index by the cooling time itself, the
+                # accumulation up to the previous time must be subtracted from
+                # time.cooling_time
+                else:
+                    actual_cooling_time = time.cooling_time - previous_times
+                
+                previous_times = (
+                    time.cooling_time if time.cooling_time != 0 else 0.0
+                )
 
-                row['time'] = time.cooling_time
+                row['time'] = actual_cooling_time
                 for n in time.nuclides:
                     nuclide = f'{n.element.lower()}-{n.isotope}{n.state}'
                     all_nucs.update({nuclide})


gonuke:
I'm a little confused by this.  Can you provide an example of what this looks like using some numerical values?

eitan-weinstein:
Sure, here's an example of how this plays out. I've printed out side by side the cooling times from an ADF with an ALARA run and one from a FISPACT-II run, without this update. The cooling times look like this:

| ALARA (s)  | FISPACT-II (s)|
|--------|------------|
| 0      | 0          |
|315.36 | 315.58 |
|938.62 | 1254.84|
|2793.67 | 4050.44|
|...|...|
|3.15e11|1.256e12|


Looking at the first time after shutdown, there is agreement between ALARA and FISPACT-II's values, but by the second cooling time, what ALARA reports as 938.62 s, the FISPACT-II cooling time shows 1254.84 s. Notably, this differs from the listed value in the actual FISPACT-II output file at this cooling time, which is 939.26 s. Rather, the value that `pypact` reports as the cooling time here is the sum of all cooling times up to that point. Subtracting the first cooling time value from the second in the FISPACT-II list yields `1254.84 s - 315.58 s = 939.26 s`, which is the exact value recorded in the FISPACT-II output file and roughly matches the corresponding ALARA value.

To formulate this, let _t<sub>i</sub>_ be the actual cooling time at index _i_, as reported in a FISPACT-II output file and _T<sub>i</sub>_ be the reported cooling time that `pypact` calculates. In order to get the desired value of the actual cooling time, the previous `pypact` "cooling time" must be subtracted from the current.

_T<sub>i</sub>_ = $\sum_{j=0}^{i}$_t<sub>j</sub>_
_t<sub>i</sub>_ = _T<sub>i</sub>_ - _T<sub>i-1</sub>_ 

Naturally, this discrepancy grows cumulatively, such that the final cooling time without this fix in the example series in reference here is ~4x larger than actual value.


Applying the updates from this PR, the same side-by-side comparison looks like this, with a mean relative difference of 6.85e-4 and a maximum relative difference of 7.26e-2, corresponding simply to the small differences in how ALARA and FISPACT-II report the precision of the cooling times:


| ALARA (s)  | FISPACT-II (s)|
|--------|------------|
| 0      | 0          |
|315.36 | 315.58 |
|938.62 | 939.26|
|2793.67 | 2795.6|
|...|...|
|3.15e11|3.15e11|

eitan-weinstein:
This update also clarifies some of the discrepancies that we've been seeing in the 10-100 year activation responses between ALARA and FISPACT-II, which have been independent from the ALARAJOY updates. For example, looking at the comparison of total decay heat responses  for ALARA, ALARAJOY/TENDL-2017 and ALARA, EAF-2010 against FISPACT-II, TENDL-2017 (all with UKDD-2020 decay data) from before this update, there appears to be almost no actual discrepancy between the ALARA and FISPACT-II totals in the plot on the left showing the time series of the responses themselves, until about ~5e9 s after shutdown. However, looking at the plot on the right of the ratios of each ALARA run against FISPACT-II, there is clearly a period in which ALARA calculates a much higher (up to 4 orders of magnitude) decay heat response than FISPACT-II, which is shared for both ALARAJOY/TENDL-2017 and EAF-2010. Such a large discrepancy _should_ be visible on the plot on the left of the non-relative values, but it does not. Moreover, the plot on the left shows the FISPACT-II time series extending noticeably further than either ALARA run, despite them having the same irradiation parameters (corresponding to the increasingly large cooling time misrepresentations discussed above at late cooling times). 
Iron totals

After applying this update, however, we produce total response plots that may show less agreement between ALARA and FISPACT-II, but are internally consistent between the plots of the actual values and the ratio plots unlike before. Now, this ALARA/FISPACT-II discrepancy is visible in both plots, as it should be.

image

The ratio time series arrays are calculated by dividing the response array of a sample run against the control run (in this case the control being FISPACT-II). Given that the length of the sample and control arrays were the same and the response values in them were correct for the cooling times they actually corresponded to, the ratio plots were accurately depicting this discrepancy between ALARA and FISPACT-II. This can be seen the fact that, other than the series' colors being different for the two ratio plots above, the time series themselves are identical to each other because they depended only on the indexing of the FISPACT-II cooling times, not the times themselves. For the non-ratio plots, however, the FISPACT-II values were not mapping on to the appropriate corresponding cooling times. That we consistently produced total response plots across like above that _appeared_ to be in strong agreement seems to have been a bizarre and misleading fluke in the way ALARA systemically calculates higher decay responses at these time orders which were effectively cancelled out by the values of increasingly over-stretched FISPACT-II cooling times.

I'm very glad that I uncovered this for a number of reasons, not least because I've been perplexed by why we've seen such dramatic spikes across all single-element irradiations in the "Activation Handbook" study when the plotted raw data were not showing the same behavior. I was partially reassured that these spikes were consistent across ALARA runs and did not appear to be exacerbated or modified in any significant way by the ALARAJOY updates, making this issue somewhat secondary in my concerns about engineering a data pipeline that produced inventory results consistent with how we would expect ALARA to behave and with fidelity to the original TENDL data.

Nevertheless, with this clarification/reaffirmation that these discrepancies are (a) real and (b) responses to systemic differences in how ALARA and FISPACT-II calculate their results, it amplifies the already existing need to find a meaningful, quantifiable metric by which to show that our activation response is indeed faithful to the original nuclear data, even though the responses themselves may not be entirely similar. Given that this PR now gives us a clearer sense that we should not be expecting the medium-term (10-100 y) responses to match between ALARA and FISPACT-II, I need to figure out how best to demonstrate that the extent and temporal distribution of the differences are not due to the ALARAJOY methods and would be interested in discussing this further. While I would also like to better understand the fundamental reasons for why ALARA and FISPACT-II are producing these differences, a thorough interrogation of that may be outside the scope of our demonstration of the fidelity of the ALARAJOY updates for the publication and the project in general.

eitan-weinstein:
Following up on this, I think that the pairing of this PR with #279 to be able to effectively run TENDL-2014 through ALARAJOY will be able to give us the necessary normalization to comment on the validity of the ALARAJOY processing methods for this segment of the study. While we are expecting some consistent differences between ALARA and FISPACT-II at these time scales regardless of the input cross-section library, the relative differences between runs from the same inventory software but different TENDL releases should be consistent if ALARAJOY is working appropriately. In other words, we expect there to be some differences between TENDL-2014 and TENDL-2017 data (given that they continue to produce new and updated versions) and we expect there to be some systemic differences in the decay responses between ALARA and FISPACT-II. But across ALARA or FISPACT-II, the relative differences between the two TENDL versions should be equivalent.

For example, looking at the plot below for the specific activity response for the elemental iron irradiation, the blue-dashed line corresponds to ALARA's ratio-time series of TENDL-2017 against TENDL-2014, both of which were processed through ALARAJOY. The orange dash-dot line corresponds to the same for FISPACT-II (with whatever closed-source data processing methods they apply to get their groupwise cross-sections). While there is still a divergence in these series, which matches the regions in the plots in earlier comments, the mean of the ratios for each time series is very close to one another (0.965 for ALARA, 0.944 for FISPACT-II), corresponding to a percentage difference of 2.2%. 

image


Ι think that this type of value will be useful in quantitatively summarizing each element's overall performance (though for high percentage differences, a closer look at specific nuclides/reactions will be warranted). I found this plotting tool that I think will be helpful in representing this data: https://github.com/Andrew-S-Rosen/periodic_trends and will go through and start re-compiling these results.

gonuke:
This whole thread makes me suspicious that ALARA and FISPACT specify cooling times in different ways.  Is it possible that ALARA specifies the absolute cooling times from shutdown and that FISPACT specifies the increment in cooling times from the previous cooling time?  That would explain my pypact is summing the cooling times, because they are actually cumulative times.  It also means that we should be specifying different cooling times in ALARA if we want to have corresponding times.

I don't know anything about FISPACT, but it's the only reason that it would make sense for pypact to be adding the times together.

I fear that we may be trying to address this discrepancy in the exact opposite way than what we should?

eitan-weinstein:
I don't think that FISPACT-II and ALARA are reporting cooling times differently, except for the pre-irradiation time and shutdown itself, but I think that we've already taken care of those discrepancies when I originally introduced FISPACT-II output parsing to the ADF framework in #231. For example, I did a simple simulation with cooling times of 1,2, and 3 years, and the reported cooling times in the FISPACT-II output file reported the absolute cooling times, rather than 1 year for each, which would be the case if it was just reporting the increment.

gonuke:
So, do you have evidence that PyPact is reporting different times than the FISPACT output file?

gonuke:
All the evidence you have reported seems to be empirical evidence that this is what you need to do to get ALARA to match FISPACT, which makes me nervous.

I'd rather see evidence that compares the information in a FISPACT output file to the information reported by PyPact and notes a discrepancy there.

eitan-weinstein:
Sure, here's a comparison of a list of FISPACT-II's actual time values for an irradiation where I set the cooling times to be every minute for 10 minutes after shutdown against `pypact`'s reported values (2 year irradiation = 6.3115E+07 s):

 | FISPACT-II (s) | pypact (s)| Accumulation|
|--------|------------|------------|
| 0 (pre-irradiation)     | 0 (pre-irradiation)|`=0+0`|
| 6.3115E+07 (shutdown) | 0 (shutdown) |Separate `pypact` method for adjusting for shutdown|
|60| 60|`=0+60`|
|120| 180|`=60+120`|
|180|360|`=180+180`|
|240 |600|`=360+240`|
|300 |900|`=600+300`|
|360 |1260|`=900+360`|
|420 |1680|`=1260+420`|
|480 |2160|`=1680+480`|
|540 |2700|`=2160+540`|
|600 |3300|`=2700+600`|

As I mentioned in the original description, the `pypact` cooling times are intentionally modifying the original values from the FISPACT-II output files with this line of code:

` self.cooling_times = list(accumulate(self.cooling_times))`

The use of the `itertools.accumulate()` function accomplishes the following (from the `itertools` docs):

> Make an iterator that returns accumulated sums or accumulated results from other binary functions.
The function defaults to addition. The function should accept two arguments, an accumulated total and a value from the iterable.
If an initial value is provided, the accumulation will start with that value and the output will have one more element than the input iterable.

So my point is that I don't think this has anything directly to do with matching with ALARA, but rather to undo the accumulation that `pypact` automatically does to change the cooling times in the first place to have values that match the original cooling times from FISPACT-II itself.

eitan-weinstein:
In the end, it turns out that FISPACT-II's input for cooling times is looking for the time _intervals_, as we came to understand in our meeting today, meaning that `pypact` is correctly accumulating these values to get actual cumulative cooling times. These should match with ALARA, pending appropriate updates to my FISPACT-II input files. Closing this PR.

PR #278, opened 2026-05-18
alaraplot shading for dominant nuclide contributions to single response plots

eitan-weinstein:
This PR introduces new functionality to `tools/alara_output_processing/alara_output_plotting.plot_single_response()` to shade regions of a decay response vs time plot to indicate the dominant nuclide contributor at those times and the relative dominance of said contribution. An example for the comparison of the Total Specific Activity vs Time for the irradiation of elemental iron between ALARA with TENDL-2017 cross-sections, ALARA with EAF-2010 cross-sections, and FISPACT-II with TENDL-2017 cross-sections looks like this:

image
image

The darkness of the shading of a region corresponds to the proportion of the total response contributed by the dominant nuclide, with a higher proportion leading to a less transparent shading. Furthermore, if multiple runs are plotted comparatively, the shading is cumulative, so darker regions can also indicate agreement between runs.

I figured an addition like this would be helpful to include in our plotting module, especially to make "total" plots more informative in showing which nuclides are most prominent across cooling times.

### DIFF
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -657,7 +657,8 @@ def plot_single_response(
     control_run=None,
     sig_figs=3,
     mark_thalf=False,
-    shading=False
+    shading=False,
+    shading_color_map={}
 ):
     '''
     Create a simple x-y plot of a given variable tracked in an ALARA output
@@ -740,6 +741,10 @@ def plot_single_response(
             particular nuclide is the dominant contributor to the plotted
             decay response.
             (Defaults to False)
+        shading_color_map (dict, optional): Option to import a pre-existing
+            nuclide color map to shade dominant regions consistently with
+            other plots.
+            (Defaults to {})
 
     Returns:
         fig (matplotlib.figure.Figure): Closed Matplotlib Figure object
@@ -747,6 +752,8 @@ def plot_single_response(
         legend_fig (matplotlib.figure.Figure or None): Conditionally separated
             Matplotlib Figure object containing only the plot's legend. None
             if separate_legend argument is False.
+        shading_color_map (dict): Color map for dominant nuclide shaded
+            regions. Only populated if shading=True.
     '''
 
     ratio_plotting = (control_run is not None)
@@ -802,8 +809,8 @@ def plot_single_response(
         cmap_name='Reds', pivs=pivs, mark_thalf=mark_thalf
     )
 
-    shading_color_map = {}
-    if shading:
+
+    if shading and not shading_color_map:
         shading_color_map = build_color_map(
             cmap_name=cmap_name, pivs=list(shade_pivs.values())
         )
@@ -931,7 +938,7 @@ def plot_single_response(
     ax.grid(True)
     plt.tight_layout(rect=[0, 0, 0.85, 1])
 
-    return fig, legend_fig
+    return fig, legend_fig, shading_color_map
 
 def single_time_pie_chart(
     agg,
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -568,19 +568,23 @@ def shade_dominant_nuclides(piv, ax, color_map, cmap_name):
     # Calculate logarithmic half-way values for each cooling time for shading
     # to flow smoothly between nuclide regions
     logt = np.log10(times)
-    bounds = np.empty(len(times) + 1)
-    bounds[1:-1] = 10 ** ((logt[:-1] + logt[1:]) / 2)
-    bounds[0]    = 10 ** (logt[0] - (logt[1] - logt[0]) / 2)
-    bounds[-1]   = 10 ** (logt[-1] + (logt[-1] - logt[-2]) / 2)
+    half_delta_logt = 0.5 * np.diff(logt)
+    log_bounds = np.empty(len(times) + 1)
+    log_bounds[0] = logt[0] - half_delta_logt[0]
+    log_bounds[1:-1] = logt[:-1] + half_delta_logt
+    log_bounds[-1] = logt[-1] + half_delta_logt[-1]
+    bounds = np.nan_to_num(10**log_bounds, nan=0.0)
 
     # Calculate the time bounds for each dominant nuclide's period of leading
     # contribution to the response variable
     dominance_ranges = {}
     for i, nuc in enumerate(dominant_nucs):
         if nuc not in dominance_ranges:
-            dominance_ranges[nuc] = [bounds[i], bounds[i + 1]]
+            dominance_ranges[nuc] = [[bounds[i], bounds[i + 1]]]
+        elif dominance_ranges[nuc][-1][1] == bounds[i]:
+            dominance_ranges[nuc][-1][1] = bounds[i + 1]
         else:
-            dominance_ranges[nuc][1] = bounds[i + 1]
+            dominance_ranges[nuc].append(bounds[i], bounds[1 + i])
 
     for i, (nuc, dominance) in enumerate(zip(dominant_nucs, relative_max)):
         ax.axvspan(
@@ -815,8 +819,10 @@ def plot_single_response(
             ax, shading_color_map, dominance_ranges = shade_dominant_nuclides(
                 shade_piv, ax, shading_color_map, cmap_name=cmap_name
             )
-            for nuc, (tmin, tmax) in dominance_ranges.items():
-                all_dominance_ranges[nuc].append((run_lbl, tmin, tmax))
+
+            for nuc, ranges in dominance_ranges.items():
+                for tmin, tmax in ranges:
+                    all_dominance_ranges[nuc].append((run_lbl, tmin, tmax))
 
         for nuc in piv.index:
             if nuc == 'total' and not total:
@@ -845,10 +851,15 @@ def plot_single_response(
                         f'\\ \\sigma = {y.std():.{sig_figs}g}$\n――――――'
                     )
 
+            x = piv.columns
+            last_nonzero_time = x[np.flatnonzero(y)[-1]]
+            if last_nonzero_time > xmax:
+                xmax = last_nonzero_time
+
             plot_or_scatter(
                 ax=ax,
                 plot_type=plot_type,
-                x=piv.columns,
+                x=x,
                 y=y,
                 label=(nuc + label_suffix),
                 color=color_map[nuc] if nuc != 'total' else "#574949",
@@ -896,6 +907,7 @@ def plot_single_response(
     ax.set_ylabel(ylabel)
     ax.set_xlabel(f'Time ({time_unit})')
     ax.set_xscale('log')
+    ax.set_xlim(right=xmax)
     ax.set_yscale(yscale)
     if ymin:
         ax.set_ylim(bottom=ymin)
--- a/tools/alara_output_processing/alara_output_plotting.py
+++ b/tools/alara_output_processing/alara_output_plotting.py
@@ -774,6 +774,7 @@ def plot_single_response(
         run_lbls = [run_lbls]
 
     data_list = []
+    shade_pivs = {}
     styles = list(
         lines.lineStyles.keys() if plot_type == 'plot'
         else lines.lineMarkers.keys()
@@ -795,22 +796,7 @@ def plot_single_response(
             control_piv = piv
         else:
             data_list.append((run_lbl, filtered, piv, style))
-
-    pivs = [data[2] for data in data_list]
-    color_map = build_color_map(cmap_name=cmap_name, pivs=pivs)
-    thalf_cmap = build_color_map(
-        cmap_name='Reds', pivs=pivs, mark_thalf=mark_thalf
-    )
-
-    plotted_nucs = []
-    xmax = 0
-    all_dominance_ranges = defaultdict(list)
-    shading_color_map = {} if total else color_map
-
-    for run_lbl, filtered, piv, style in data_list:
-        if shading:
-            shade_piv = piv
-            if total:
+            if shading and total:
                 _, shade_piv = preprocess_data(
                     adf=adf,
                     run_lbl=run_lbl,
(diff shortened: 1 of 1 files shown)


tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -511,6 +517,131 @@ def add_pie(ax, time_slice, color_map, threshold):
...
+            cmap_name=cmap_name, all_nucs=set(dominant_nucs)
+        )
+
+    # Calculate logarithmic half-way values for each cooling time for shading
+    # to flow smoothly between nuclide regions
+    logt = np.log10(times)
+    bounds = np.empty(len(times) + 1)
+    bounds[1:-1] = 10 ** ((logt[:-1] + logt[1:]) / 2)
+    bounds[0]    = 10 ** (logt[0] - (logt[1] - logt[0]) / 2)
+    bounds[-1]   = 10 ** (logt[-1] + (logt[-1] - logt[-2]) / 2)
```
gonuke: For clarity, let's assemble `log_bounds` first and then raise it all to the power. That might make it more clear how you're doing the min/max - something like:

```suggestion
    half_delta_logt = 0.5 * (logt[1:] - logt[:-1])
    log_bounds = np.empty(len(times) + 1)
    log_bounds[0] = logt[0] - half_delta_logt[0]
    log_bounds[1:-1] = logt[:-1]+ half_delta_logt
    log_bounds[-1] = logt[-1] + half_delta_logt[-1]
    bounds = 10 ** log_bounds
```

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -511,6 +517,131 @@ def add_pie(ax, time_slice, color_map, threshold):
...
+
+    # Calculate the time bounds for each dominant nuclide's period of leading
+    # contribution to the response variable
+    dominance_ranges = {}
+    for i, nuc in enumerate(dominant_nucs):
```
gonuke: Is it possible for a nuclide to have two disjoint ranges where it's dominant, and will this allow that?

gonuke (changes requested):
This looks like a nice features - just a couple of questions for now...

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: Good catch. It would definitely be possible for a nuclide to have disjoint dominance ranges, especially in pulsed irradiations. I don't think, though, that my current implementation would allow for that, so I will work on updating it to be able to handle such cases.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -511,6 +517,140 @@ def add_pie(ax, time_slice, color_map, threshold):
...
+        if nuc not in dominance_ranges:
+            dominance_ranges[nuc] = [[bounds[i], bounds[i + 1]]]
+        elif dominance_ranges[nuc][-1][1] == bounds[i]:
+            dominance_ranges[nuc][-1][1] = bounds[i + 1]
+        else:
+            dominance_ranges[nuc].append(bounds[i], bounds[1 + i])
```
gonuke: Shouldn't this be appending a list:
```suggestion
            dominance_ranges[nuc].append([bounds[i], bounds[1 + i]])
```

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -511,6 +517,140 @@ def add_pie(ax, time_slice, color_map, threshold):
...
+
+    for i, (nuc, dominance) in enumerate(zip(dominant_nucs, relative_max)):
+        for rnge in dominance_ranges[nuc]:
+            if np.diff(rnge) > 0:
```
gonuke: When does this fail? Can't we prevent it by more carefully setting the ranges?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -511,6 +517,140 @@ def add_pie(ax, time_slice, color_map, threshold):
...
+
+    # Calculate logarithmic half-way values for each cooling time for shading
+    # to flow smoothly between nuclide regions
+    logt = np.log10(times)
+    half_delta_logt = 0.5 * np.diff(logt)
+    log_bounds = np.empty(len(times) + 1)
+    log_bounds[0] = logt[0] - half_delta_logt[0]
+    log_bounds[1:-1] = logt[:-1] + half_delta_logt
+    log_bounds[-1] = logt[-1] + half_delta_logt[-1]
+    bounds = np.nan_to_num(10**log_bounds, nan=0.0)
```
gonuke: Also just realized that this is the same as:
```suggestion
    bounds = np.empty(len(times) + 1)
    bounds[1:-1] = np.sqrt(times[1:] * times[:-1])
    bounds[0] = times[0] * times[0]/bounds[1]
    bounds[-1] = times[-1] * times[-1]/bounds[-2]
```

gonuke (changes requested):

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: I think this was failing because of the lack of brackets in the appending above, per your last comment. This shouldn't be needed anymore.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -511,6 +517,134 @@ def add_pie(ax, time_slice, color_map, threshold):
...
+        elif dominance_ranges[nuc][-1][1] == bounds[i]:
+            dominance_ranges[nuc][-1][1] = bounds[i + 1]
+        else:
+            dominance_ranges[nuc].append([bounds[i], bounds[1 + i]])
+
+    for i, (nuc, dominance) in enumerate(zip(dominant_nucs, relative_max)):
+        for _ in dominance_ranges[nuc]:
+            ax.axvspan(
+                bounds[i],
+                bounds[i + 1],
```
gonuke: This doesn't seem to take advantage of the work you did to build these dominance ranges, especially since you just throw them away.  I think these data structures are more complicated than they need to be.

Related, I'm not sure how you assign `alpha` when you have multiple adjacent regions that are dominated by the same nuclide, but with different values of `dominance`.  If they are treated as separate regions, then maybe you don't actually need to build the `dominance_ranges` at all?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -511,6 +517,134 @@ def add_pie(ax, time_slice, color_map, threshold):
...
+                color=color_map[nuc],
+                # Shading transparency as an inverse function of the relative
+                # contribution of the dominant nuclide
+                alpha=0.2 * dominance,
+                linewidth=0,
+                label=None
+            )
```
gonuke: I don't think you need to care about `dominance_ranges` in this loop - I'd recommend:


```suggestion
    for lower, upper, nuc, dominance in zip(bounds[:-1], bounds[1:], dominanct_nucs, relative_max):
        ax.axvspan(
            lower,
            upper,
            color=color_map[nuc],
            # Shading transparency as an inverse function of the relative
            # contribution of the dominant nuclide
            alpha=0.2 * dominance,
            linewidth=0,
            label=None
        )
```

gonuke (changes requested):
Just trying to simplify things a little more.

Also - is it too much to plot dominance ranges for multiple run labels on the same plot?  They overlap each other and are hard to see.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -511,6 +517,121 @@ def add_pie(ax, time_slice, color_map, threshold):
...
+    Returns:
+        None 
+    '''
+
+    for nuc, run_entries in all_dominance_ranges.items():
+        run_lines = '\n'.join(
+            f'   -  {rl}: {tmin:.2g} - {tmax:.2g} {time_unit}'
+            for rl, tmin, tmax in run_entries
+        )
+        ax.axvspan(
```
gonuke, line 630: I think this is making zero-width spans just to generate a legend label.  It might be useful to say that in a comment somewhere.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -650,15 +777,52 @@ def plot_single_response(
...
+                shade_pivs[run_lbl], ax,
+                shading_color_map, cmap_name=cmap_name
+            )
+
+            for nuc in set(dominant_nucs):
+                indices = [i for i, n in enumerate(dominant_nucs) if n == nuc]
+                all_dominance_ranges[nuc].append((
+                    run_lbl,
+                    bounds[indices[0]],     # t_min
+                    bounds[indices[-1] + 1] # t_max
```
gonuke: I think this is not handling disjoint ranges again.  Indices should always have an even number of entries where `indices[::2]` are the lower bounds of each range and `indices[1::2]` are the upper bounds.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -686,23 +850,34 @@ def plot_single_response(
                         f'\\ \\sigma = {y.std():.{sig_figs}g}$\n――――――'
                     )
 
+            x = piv.columns
```
gonuke: I think `t` is a better choice here than `x` but I see that you have used `y` above....  In general, it's better for variables to semantically evoke their physical interpretation rather than their plot axis.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -686,23 +850,34 @@ def plot_single_response(
...
+
             plot_or_scatter(
                 ax=ax,
                 plot_type=plot_type,
-                x=piv.columns,
+                x=x,
                 y=y,
                 label=(nuc + label_suffix),
-                color=color_map[nuc],
+                color=color_map[nuc] if nuc != 'total' else "#574949",
```
gonuke: I haven't looked carefully, but couldn't you just inject this into the `color_map` as the value for `total` when you build the color map?

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -511,6 +517,121 @@ def add_pie(ax, time_slice, color_map, threshold):
...
+            # Shading transparency as an inverse function of the relative
+            # contribution of the dominant nuclide
+            alpha=0.2 * dominance,
+            linewidth=0,
+            label=None
+        )
+
+    return ax, bounds, dominant_nucs
+
+def collect_shading_labels(ax, color_map, all_dominance_ranges, time_unit):
```
gonuke: Maybe a better function name is something like...
```suggestion
def add_shading_legend_labels(ax, color_map, all_dominance_ranges, time_unit):
```

It doesn't return anything, but adds the label to the legends (in)directly

gonuke (changes requested):
A few final notes here - we're getting close

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: You're right, although applying that kind of slicing will treat contiguous segments as disjointed as well which appears in the legend as many discrete ranges. In my new commit, I think my approach should handle this and allow for disjoint ranges to be captured correctly.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -650,15 +778,61 @@ def plot_single_response(
...
+                        contiguous_ranges.append(range_start, prev)
+                        range_start = curr
+                contiguous_ranges.append((range_start, indices[-1]))
+
+                for lower, upper in contiguous_ranges:
+                    all_dominance_ranges[nuc].append((
+                        run_lbl,
+                        bounds[lower],
+                        bounds[upper + 1]
+                    ))
```
gonuke: This is probably simpler if you loop over the bounds
```suggestion
            prev_nuc = dominant_nucs[0]
            lower = bounds[0]
            for upper, nuc in zip(bounds[1:],dominant_nucs[1:]):
                if nuc != prev_nuc:
                   all_dominance_ranges[prev_nuc].append(
                       (run_lbl, lower, upper)
                   )
                   lower = upper
                   prev_nuc = nuc
               all_dominance_ranges[prev_nuc].append(
                   (run_lbl, lower, upper)
               )           
```

gonuke (changes requested):
A simpler option for the disjoint ranges

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: Happy to change `x` to `t`. I could just change `y` to `series` or `f_t` if you think that would be a more appropriate pairing.

tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -650,15 +781,56 @@ def plot_single_response(
...
+            lower = bounds[0]
+            for upper, nuc in zip(bounds[1:], dominant_nucs[1:]):
+                if nuc != prev_nuc:
+                    all_dominance_ranges[prev_nuc].append((
+                        run_lbl, lower, upper
+                    ))
+                    lower = upper
+                    prev_nuc = nuc
+
+            all_dominance_ranges[prev_nuc].append((run_lbl, lower, upper))
```
gonuke: I think I introduced a bug - maybe you can check?
```suggestion
            for upper, nuc in zip(bounds[1:-1], dominant_nucs[1:]):
                if nuc != prev_nuc:
                    all_dominance_ranges[prev_nuc].append((
                        run_lbl, lower, upper
                    ))
                    lower = upper
                    prev_nuc = nuc
            upper = bounds[-1]
            all_dominance_ranges[prev_nuc].append((run_lbl, lower, upper))
```

gonuke (changes requested):
I think my last suggestion had a bug...

tools/alara_output_processing/alara_output_plotting.py
eitan-weinstein: Looking at resultant plots with one version versus the other, it's not obvious that your original suggestion caused a bug, but I do think that logic is more sound with what your are offering now.

gonuke (approved):

PR #279, opened 2026-05-19
Handling TENDL-2014 processing errors in ALARAJOYWrapper

eitan-weinstein:
This PR introduces some minor updates for the PENDF production stage of ALARAJOYWrapper to allow for more robust handling in the case of a TENDL file missing necessary energy-dependent data in the unresolved region. The motivation behind this update is that I wanted to produce an ALARAJOY/TENDL-2014 library as well for my analysis, figuring that one possible solution to the question on FISPACT-II/ALARA comparisons that I posed in the discussion in #277 would be to compare ratio plots of FISPACT-II TENDL-2017/TENDL-2014 against ALARA/ALARAJOY TENDL-2017/TENDL-2014 to filter out the impacts of the fundamental differences in the two software's decay chain production. Additionally, while the primary motivation for building ALARAJOYWrapper is to make ALARA compatible specifically with FENDL3.2x (TENDL-2017 cross-sections), it would be undoubtedly valuable if this tool could be used for any TENDL release. 

When running ALARAJOYWrapper with the [TENDL-2014 distribution from IAEA](https://www-nds.iaea.org/public/download-endf/TENDL-2014/n/), I ran into two issues. First, for a few files, UNRESR was failing with this NJOY error message:

```
***error in rdunf2***energy dependent data undefined
                      between  8.9001E+03 and  8.9001E+03 eV
```

This message indicates a deprecation of the data as the NJOY manual describes in section 5.6:

>_When using unresolved resonance formalisms with energy dependent parameters (e.g. the fission width, etc.), these data need to be defined over the entire unresolved resonance region. In rare cases, such as ENDF/BVII.0 Pu239, this is not the case, leading to NaN cross section values. This is an evaluation error._

While running UNRESR is generally preferable and part of our canonical methods for ALARAJOYWrapper, I think an evaluation error of this sort is not resolvable by us. Even without UNRESR, however, GASPR and GROUPR can still be run and the best possible data can be extracted from any given file with this issue.

A second issue in the TENDL-2014 data is that there are a number of files that are misnamed, such as the files for <sup>285</sup>Nh and <sup>286</sup>Nh, which for some reason have the symbol A3 instead. This update will catch this type of issue, identify the proper nuclide identifiers from the file itself and then rename it to match the correct nuclide data.

Finally, for a number of TENDL-2014 nuclides the RECONR error tolerance that was otherwise a fixed value was leading to NJOY stalling out and unable to resolve the reconstructions. I've included a looping mechanism to test out various timeout lengths before increasing this tolerance value to try to isolate the need to raise it only to cases where NJOY is truly unable to perform the calculation, rather than just doing so slowly.

### DIFF
--- a/tools/ALARAJOYWrapper/preprocess_fendl3.py
+++ b/tools/ALARAJOYWrapper/preprocess_fendl3.py
@@ -113,9 +113,7 @@ def configure_logging(redirect_warnings=False):
 
     logger.addHandler(console_handler)
 
-def process_pendf(
-    njoy_prep_input, material_id, MTs, pKZA, mt_dict, temperature, tendl_path
-):
+def process_pendf(material_id, MTs, pKZA, mt_dict, temperature, tendl_path):
     """
     Prepare and run initial NJOY run with MODER, RECONR, BROADR, UNRESR, and
         GASPR modules to prodece the requisite PENDF file for a subsequent
@@ -151,11 +149,36 @@ def process_pendf(
 
     element, A = tp.interpret_KZA(pKZA)
     njoy_error = ''
-    njoy_input = njt.fill_input_template(
-        njoy_prep_input, material_id, MTs, element, A, mt_dict, temperature
+
+    # Run MODER, RECONR, BROADR, UNRESR
+    njoy_prep_input = njt.fill_input_template(
+        njt.njoy_prep_input, material_id, MTs,
+        element, A, mt_dict, temperature
     )
-    njt.write_njoy_input_file(njoy_input)
-    pendf_path, njoy_error = njt.run_njoy(element, A, material_id, 'PENDF')
+    njt.write_njoy_input_file(njoy_prep_input)
+    pendf_path, prep_error, njoy_out = njt.run_njoy(
+        element, A, material_id, 'PENDF'
+    )
+
+    # If UNRESR fails, run GASPR with PENDF output of BROADR with warning
+    unresr_error_flag = '***error in rdunf2***'
+    if prep_error and unresr_error_flag in njoy_out:
+        warnings.warn(
+            'UNRESR failed to produce cross-sections in the unresolved '\
+            f'energy range for {element}{A}:\n'\
+            f'{njoy_out[njoy_out.find(unresr_error_flag):]}Skipping to GASPR.'
+        )
+        gaspr_input = njt.fill_input_template(
+            njt.gaspr_input, material_id, MTs, element, A,
+            mt_dict, temperature, unresr_fail=True
+        )
+        njt.write_njoy_input_file(gaspr_input)
+        pendf_path, gaspr_error, _ = njt.run_njoy(
+            element, A, material_id, 'PENDF'
+        )
+        njoy_error += gaspr_error
+    else:
+        njoy_error += prep_error
 
     _, pendf_MTs = tp.extract_endf_specs(pendf_path)
     MTs |= pendf_MTs.intersection(set(rxd.GAS_DF['total_mt']))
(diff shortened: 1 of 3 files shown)


tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -48,24 +50,61 @@ def args():
...
+        endf_path (pathlib._local.PosixPath): Copy of the original path to the
+            selected ENDF file, unless the nuclide naming needed to be fixed,
+            in which the new filename corresponds to the updated nuclide name.
     """
 
-    Z = njt.elements[element]
+    try:
+        Z = njt.elements[element]
+
+    # Special handling for improperly labeled nuclides (i.e. Nh-285 and Nh-286
```
gonuke: This seems intrusive to this code to handle a rare case.  Can we just introduce something to the workflow that fixes the names of these files or something similar and run that before we process the data?

gonuke (changes requested):
Can you summarize how many files and (if not too many) which files are affected by this UNRESR problem?

Also see my comment about handling poorly named files with additional workflow steps rather than while we are processing them.

tools/ALARAJOYWrapper/preprocess_fendl3.py
eitan-weinstein: Per my comment in #271  about moving away from relying on filenames for cross-section data, I'm wondering if it would make more sense to avoid this entirely both because of its intrusiveness but also because the actual data contained within the file is correct, just the naming is incorrect. Ultimately, I don't think we care that much about toying with the individual files in a TENDL or decay data release unless the data themselves need to be modified/reformatted.

eitan-weinstein:
Rebased to follow #281, which should avoid the need for the special handling for misnamed files.

gonuke:
I feel like this should not need a rebase again, but it looks like it does.

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -147,25 +153,85 @@ def process_pendf(
...
+    max_err = 0.02
+    timeouts = [60, 300] # s
+    success = False
+    while err <= max_err and not success:
+        for timeout in timeouts:
+            njoy_prep_input = njt.fill_input_template(
+                njt.njoy_prep_input, material_id, MTs,
+                element, A, mt_dict, temperature, err=err
+            )
+            njt.write_njoy_input_file(njoy_prep_input)
```
gonuke: These lines don't depend on the `timeout` value - can we put them before the timeout loop?

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -147,25 +153,85 @@ def process_pendf(
...
+    # Run MODER, RECONR, BROADR, UNRESR
+    # If fractional error tolerance to low for RECONR, can cause NJOY to time
+    # out. If this occurs, increase error tolerance until NJOY can run
+    # successfully.
+    err = 0.001
+    max_err = 0.02
+    timeouts = [60, 300] # s
+    success = False
+    while err <= max_err and not success:
+        for timeout in timeouts:
```
gonuke: I'm not sure what the benefit is of trying both timeout values.  If it makes it through the first time (60 s), presumably it would end just as quickly with the 300 s timeout.  If it doesn't make it through the first time, then we've wasted 60 s....

gonuke (changes requested):
One cleanup/streamline

gonuke (approved):
Thanks for these improvements @eitan-weinstein

PR #280, opened 2026-05-19
Enabling warning redirection for ALARAJOYWrapper

eitan-weinstein:
This PR introduces a small update to ALARAJOYWrapper to allow the user to redirect non-fatal warnings to a log file to avoid cluttering the terminal printouts.

### DIFF
--- a/tools/ALARAJOYWrapper/njoy_tools.py
+++ b/tools/ALARAJOYWrapper/njoy_tools.py
@@ -362,8 +362,10 @@ def run_njoy(element, A, matb, file_capture):
     }
 
     # Run NJOY
-    result = subprocess.run(['njoy'], input=open(INPUT).read(),
-                        text=True, capture_output=True)
+    with open(INPUT, 'r') as f:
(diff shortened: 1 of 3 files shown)


gonuke:
Maybe take a look at the pythons `logging` module?

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -362,8 +362,10 @@ def run_njoy(element, A, matb, file_capture):
     }
 
     # Run NJOY
-    result = subprocess.run(['njoy'], input=open(INPUT).read(),
-                        text=True, capture_output=True)
+    with open(INPUT, 'r') as f:
```
eitan-weinstein, line 365: The logger was capturing a `ResourceWarning` from the previous way we were opening the input, so this change is simply to done to resolve that because it was unnecessary to have in the first place.

gonuke (approved):
LGTM - thanks for pivoting to the logger @eitan-weinstein

PR #281, opened 2026-05-28
Extracting KZAs directly from ENDF files, independently from filename

eitan-weinstein:
This PR follows conversations from #271 and #279 to identify nuclides directly from the data contained within their ENDF files, rather than relying on them being named correctly. As it stood, relying on files to be named correctly was a rather fragile system, so this PR should provide a more stable basis by which to establish nuclide identity in the form of KZA, and then subsequently extract all other identifying characteristics (chemical symbol, mass number, isomeric state) from the KZA itself.

### DIFF
--- a/tools/ALARAJOYWrapper/tendl_processing.py
+++ b/tools/ALARAJOYWrapper/tendl_processing.py
@@ -150,7 +150,7 @@ def search_for_files(dir = Path.cwd()):
                     'Element'         :            element,
                     'Mass Number'     :                  A,
                     'pKZA'            :               pKZA,
-                    'TENDL File Path' : formatted_filename
+                    'TENDL File Path' :       new_filename
                 })
 
     return file_info
--- a/tools/ALARAJOYWrapper/tendl_processing.py
+++ b/tools/ALARAJOYWrapper/tendl_processing.py
@@ -138,12 +138,12 @@ def search_for_files(dir = Path.cwd()):
                     break
 
             else:
-                formatted_filename = dir / f'{element}{A}.tendl'
-                if formatted_filename != file:
-                    file.rename(formatted_filename)
+                new_filename = dir / f'{element}{A.split(str(0))[-1]}.tendl'
+                if new_filename != file:
+                    file.rename(new_filename)
                     warnings.warn(
                         f'Improperly named TENDL file {file} renamed to ' \
-                        f'{formatted_filename} to match nuclide ID.'
+                        f'{new_filename} to match nuclide ID.'
                     )
 
                 file_info.append({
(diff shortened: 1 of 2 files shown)


tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -22,27 +23,79 @@
...
+            states.
+    """
 
-    element = stem[:i]
-    A = stem[i:]       
+    kza = str(kza)
+    A = kza[-4:-1]
+    Z = kza[:kza.find(A)].zfill(2)
+    element = list(elements.keys())[int(Z) - 1]
+    M = int(kza[-1])
```
gonuke: Why not convert to strings after doing integer math?

```suggestion
    M = kza % 10
    ZA = kza // 10
    A = kza % 1000
    Z = kza // 1000
    element = list(elements.keys())[Z - 1]
```

tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -22,27 +23,79 @@
...
+    if M > 0:
+        A += ISOMERIC_STATES[M - 1]
```
gonuke, line 98: With above changes:

```suggestion
        A = str(A) + ISOMERIC_STATES[M - 1]
```

tools/ALARAJOYWrapper/tendl_processing.py
gonuke: Related - could we just use the integer value of Z everywhere that we currently use `element`?

tools/ALARAJOYWrapper/tendl_processing.py
```diff
@@ -22,27 +23,79 @@
     for key, arr in EXCITATION_DICT.items()
     for val in arr
 }
+ISOMERIC_STATES = 'mnopqrstuvwxyz'
 
-def get_isotope(stem):
+def create_endf_file_obj(path, MF):
```
gonuke, line 28: Reviewers note - unfortunately, Github makes it hard to review code when you simply reorder functions... doesn't mean it never can happen, but it is hard to check that the function is just moved and not also changed.

gonuke (changes requested):
I think this looks pretty good with just a few suggestions.

tools/ALARAJOYWrapper/tendl_processing.py
eitan-weinstein: Good point on the integer math. Much more intuitive. Certainly, we could use `Z` everywhere instead of `element`, although, I think I like using `element`, especially for printouts and warnings because it is easier for the user to see something like "Fe-56" rather than "26-56" as an identifier and then having to map the Z to the chemical symbol themselves.

tools/ALARAJOYWrapper/tendl_processing.py
eitan-weinstein: Good to know. For your reference, I can tell you that all I did to `create_endf_file_obj()` was just move it earlier in the module so it gets defined above where it gets first called now that I am using it in my new KZA parsing function.

gonuke (approved):

PR #283, opened 2026-06-02
Create option to show/hide time bounds for shaded regions in legend

eitan-weinstein:
Follows #278 .

This PR makes a small adjustment that I had meant to include in #278 to allow for the user to specify if they do not want to show the time ranges for shading of each nuclide in the legend. This may be useful when comparing multiple runs that all have the same ranges, which could make the legend unnecessarily cluttered.


gonuke (approved):
Seems straightforward... thanks @eitan-weinstein

PR #284, opened 2026-06-02
Calculating relative maximum for shading in a safer way

eitan-weinstein:
Follows #278 .

Previously, the relative maxima were only divided by the `total` contribution if a "total" row existed. This was problematic for cases of plotting individual nuclides, but _without_ the total. Rather, I've made it so that each maximum is always calculated as relative by dividing by the cooling time column's sum. This will not affect values which are already relative, as the sum is just 1 anyways.


gonuke (approved):
Looks ok to me... thanks @eitan-weinstein

PR #285, opened 2026-06-09
Enabling ALARAJOY to Convert to Group Structures beyond Vitamin-J

eitan-weinstein:
This PR updates the `ALARAJOYWrapper` preprocessor to allow for a user to convert TENDL cross section data to groupwise cross-sections with multi-group energy group structures beyond Vitamin-J. The motivation for this update is twofold. First, when comparing against FISPACT-II for our validation efforts, in order to minimize differences between ALARA and FISPACT-II, which is primarily designed to use the CCFE-709 group structure, it would be valuable to compare cross-sections with like-structures. FISPACT-II is capable of applying other group structures as well, but my second consideration in motivating this PR is that I think it would be valuable in general for users to be able to specify their desired group structure, without dictating that Vitamin-J is the only valid option, considering that ALARA can use other group structures already. Whether or not we choose to apply this update for the validation efforts in producing a 709 group ALARAJOY library can be decided separately, but I still think an update like this will be a generally helpful feature.

The primary update that I introduced in this PR is a new function `njoy_tools.set_group_structure()`, which takes in an optional user argument for the group structure and prepares the proper format of `GROUPR` parameters to convert to the provided cross-sections. In the docstring for that function, I specify the types and ways this can be done:

> Interpret the group_structure argument (`-g`) to define the requisite NJOY
>        parameters to convert TENDL data to the given multi-group energy
>       structure. By default, the Vitamin-J 175 group structure is used for
>        ALARAJOY preprocessing and this function will return the appropriate
>        values for NJOY to run GROUPR with these settings. If another group
>       structure is desired, however, there are three ways in which this
>        group structure can be set:
>
>       1) Provide the GROUPR `ign` key corresponding to the desired group
>        structure. NJOY has 33 built-in group structures from which GROUPR can
>        access their data, including for Vitamin-J. These can be found in
>        Section 8.18 ("Running GROUPR") of the NJOY User Manual
>        (https://github.com/njoy/NJOY2016-manual/raw/master/njoy16.pdf).
>
>        2) Provide a support file containing the explicit energy bounds of a
>        multi-group energy structure. NJOY is not limited to its pre-set
>        structures, and is capable of converting cross-sections to an
>        arbitrary group structure, so long as the energy bounds and number of
>        groups are provided. To use this option format the support file as:
>
>            BOUND_1
>            BOUND_2
>            ...
>            BOUND_N
>        
>        The order in the support file can be arbitrary because after being
>        read in, the values will be sorted to ensure compliance to NJOY's
>        expectation of ascending energy bounds for an arbitrary group
>        structure.
>
>        3) Provide a group-structure name corresponding to a key in the
>        `openmc.mgxs.GROUP_STRUCTURES` dictionary. The open-source Monte Carlo
>        neutron transport code OpenMC contains a number of multi-group energy
>        structures that may be applicable for activation studies but are not
>        included in the list of NJOY options mentioned in `(1)`, such as the
>        'CCFE-709' group structure used by FISPACT-II. These group structures
>        can be found at https://docs.openmc.org/en/stable/pythonapi/mgxs.html.

### DIFF
--- a/tools/ALARAJOYWrapper/njoy_tools.py
+++ b/tools/ALARAJOYWrapper/njoy_tools.py
@@ -164,11 +164,12 @@ def set_group_structure(group_struct_arg):
         structure is desired, however, there are three ways in which this
         group structure can be set:
 
-        1) Provide the GROUPR `ign` key corresponding to the desired group
-        structure. NJOY has 33 built-in group structures from which GROUPR can
-        access their data, including for Vitamin-J. These can be found in
-        Section 8.18 ("Running GROUPR") of the NJOY User Manual
-        (https://github.com/njoy/NJOY2016-manual/raw/master/njoy16.pdf).
+        1) Provide the GROUPR `ign` key or group name corresponding to the
+        desired group structure. NJOY has 33 built-in group structures from
+        which GROUPR can access their data, including for Vitamin-J. These can
+        be found in Section 8.18 ("Running GROUPR") of the NJOY User Manual
+        (https://github.com/njoy/NJOY2016-manual/raw/master/njoy16.pdf) or in
+        the dictionary njoy_tools.NJOY_GROUPS.
 
         2) Provide a support file containing the explicit energy bounds of a
         multi-group energy structure. NJOY is not limited to its pre-set
@@ -195,14 +196,13 @@ def set_group_structure(group_struct_arg):
         can be found at https://docs.openmc.org/en/stable/pythonapi/mgxs.html.
     
     Arguments:
-        group_struct_arg (None or list of str): Group structure argument
-            following procedures described above or None, if default
-            Vitamin-J 175 group structure settings to be applied.
+        group_struct_arg (list of str): Group structure argument following the 
+            procedures described above.
 
     Returns:
-        ign (str or int): GROUPR neutron group structure parameter. ign = 1
-            for arbitrary group structures not contained in NJOY's built-in
-            list of options.
+        ign (int): GROUPR neutron group structure parameter. ign = 1 for
+            arbitrary group structures not contained in NJOY's built-in list
+            of options.
         ngn (str): Number of groups. Will be an empty string unless ign == 1.
         egn (str): Space-joined string of all energy group bounds in
             ascending order. Will be an empty string unless ign == 1.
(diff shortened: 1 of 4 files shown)


eitan-weinstein:
Per our discussions from last week's group meetings, there are some issues that I've been running into with certain cross-sections being converted to zeroes rather than to values that match faithfully to FISPACT-II's 709 structure. For example, with the cross-sections for <sup>10</sup>B (n,Xt), 


image

From the plot above, there generally seems to be matching spectral shape for the ALARAJOY processed values, as compared to FISPACT-II, except for a shift to the right at low energies and a truncation of values at high energies. From my investigations, I believe this is being caused by a failufre to produce groupwise cross-sections in all of the energies where there _should_ be values, which then get padded by zeroes. This causes two problems. The first and more fundamental, is that we are losing cross-section values that should be there. While it is certainly possible to have zero-valued cross-sections (hence the zero-padding in the first place), looking at the unmodified spectrum and FISPACT-II's groupwise spectrum, there are obviously non-zero groups that are somehow being missed. Secondly, these missing values are disrupting the indexing of the cross-section arrays. Looking at the low energy range, the large exponential curve in all of the series shows that these groups are being correctly processed by ALARAJOY, but while they should be at the beginning of the array, instead they are being padded beforehand, thus pushing the series into the rightward bias.

In trying to diagnose the source of this issue, I've gone through and catalogued all of the group structures that lack non-zero cross-sections at all values, suspecting that this may be a shared underlying feature:

| Structure | nGroups | nonZeroes | Highest Energy (eV) |
|-|-|-|-|
| CCFE-660 | 660 | 650 | 3.00e+07|
| CCFE-1025 | 1025 | 975 | 3.00e+07|
| UKAEA-1067 | 1067 | 975 | 2.00e+08 |
| UKAEA-1102 | 1102 | 975 | 1.00e+09 |
| UKAEA-142 | 142 | 121 | 2.00e+08 |
| VITAMIN-J-42 | 42 | 40 | 5.00e+07 |
| CCFE-709 | 709 | 650 | 1.00e+09 |
| UKAEA-1102 | 1102 | 975 | 1.00e+09|

eitan-weinstein:
Looking at the `EMAX` (maximum energy) value in the original <sup>10</sup>B TENDL file, the highest evaluated energy is 2e+07 eV. Compared to the table above, all of those values are above this upper limit, so NJOY would have no reference values from which to incorporate into group calculations. While this explains why we are seeing such a large expanse of zeroes, we are left with a non-trivial question of how the NEA Gitlab grouped TENDL-2017 cross-sections which are used by FISPACT-II could possibly have non-zero cross-section values for energies above 2e+07 eV...

eitan-weinstein:
Taking a step back, I wanted to compare the cross-sections for other fusion materials-relevant reactions between ALARAJOY with 709 and 175 groups, as well as the original TENDL continuous spectra cross-sections. We're seeing a consistent transformation of the 709 group cross-sections towards higher energy values (rightward in the plotted orientations). The shapes of the spectra themselves seem generally consistent, but so too is this bias. To that end, it seems that there is some fundamental mismatch when the CCFE-709 group structure (and likely its related CCFE/UKAEA high upper energy limit structures) are applied, leading to zeros being added at the incorrect energy bin index, leading to this rightward bias.

Still fully fleshing out my diagnosis of this issue and would be happy to discuss with @gonuke about what underlies this improper conversion behavior and how to move forward to resolve this both technically and to gain a deeper understanding conceptually.

image
image
image
imageimage
image
image
image
image

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -52,7 +52,16 @@ def args():
                 arise to a log file.
         ''')
     )
-    return parser.parse_args()
+    parser.add_argument(
+        '--group_structure', '-g', required=False, nargs=1,
+        help=('''
+            Specification for group structure in which to convert TENDL cross-
+                sections. See tendl_processing.set_group_structure() for
```
gonuke:
```suggestion
                sections. See njoy_tools.set_group_structure() for
```

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -116,6 +118,143 @@ def set_directory():
...
+        group_name = 'VITAMIN-J-175'
+
+    else:
+        group_struct = group_struct_arg[0]
+
+        # Check if provided group structure is among the list of built-in NJOY
+        # group structures (ign values 2-34)
+        if group_struct in np.array(list(NJOY_GROUPS)).astype(str):
+            ign = group_struct
+            group_name = NJOY_GROUPS[int(group_struct)]
```
gonuke: Why not always make this an `int`?
```suggestion
            ign = int(group_struct)
            group_name = NJOY_GROUPS[ign]
```

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -116,6 +118,143 @@ def set_directory():
...
+        'CCFE-709' group structure used by FISPACT-II. These group structures
+        can be found at https://docs.openmc.org/en/stable/pythonapi/mgxs.html.
+    
+    Arguments:
+        group_struct_arg (None or list of str): Group structure argument
+            following procedures described above or None, if default
+            Vitamin-J 175 group structure settings to be applied.
+
+    Returns:
+        ign (str or int): GROUPR neutron group structure parameter. ign = 1
```
gonuke: I think with my suggestion below, this can always be a `int`?

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -116,6 +118,143 @@ def set_directory():
...
+    28 :      'RAHAB-AECL-89',
+    29 :           'CCFE-660',
+    30 :         'UKAEA-1025',
+    31 :         'UKAEA-1067',
+    32 :         'UKAEA-1102',
+    33 :          'UKAEA-142',
+    34 :           'LANL-618'
+}
+
+def set_group_structure(group_struct_arg):
```
gonuke, line 157: Why not the following to skip the first conditional below?

```suggestion
def set_group_structure(group_struct_arg='17'):
```

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -116,6 +118,143 @@ def set_directory():
...
+    if group_struct_arg is None:
+        ign = 17
+        group_name = 'VITAMIN-J-175'
+
+    else:
+        group_struct = group_struct_arg[0]
+
+        # Check if provided group structure is among the list of built-in NJOY
+        # group structures (ign values 2-34)
+        if group_struct in np.array(list(NJOY_GROUPS)).astype(str):
```
gonuke: It might be nice if users could provide the name of a spectrum rather than just the number?

gonuke (changes requested):
This looks like it's well done.  I have a few small comments here.

tools/ALARAJOYWrapper/njoy_tools.py
eitan-weinstein: I think setting the default value for the `-g` argument as `['17']` may actually be even cleaner.

eitan-weinstein:
I think I've solved this rightward shift issue now. Previously, we were assuming that every GENDF file would list cross-sections for all groups up to the highest group with a non-zero value. Instead, each group with a non-zero cross section is listed in pairs of lines containing the following data:

```
... Group Index (IG) ...
... Cross-Section  ...
```

By extracting cross-sections from the GENDF file into a dictionary keyed by the actual group index and then populating them to an `np.zeros(nGroups)` array by the index of the actual non-zero groups, we can avoid making assumptions of positionality and instead rely on the explicit information provided within the files themselves.

From some initial plots, this is looking reassuring, such as with <sup>28</sup>Si (n,p) plot below. More analysis to follow...

image

eitan-weinstein:
This PR should be in pretty good shape here and in order to simplify some of the cross-section plotting PRs that are on the way, I think it's worth circling back for one more review so we can get it merged and build those features off from here.

tools/ALARAJOYWrapper/preprocess_fendl3.py
```diff
@@ -424,11 +450,15 @@ def main():
     # Handle gas total production cross-sections, per user specifications
     gas_filtered = subtract_gas_from_totals(all_rxns)
 
-    if args().amalgamate:
-        gas_filtered = combine_daughter_pathways(gas_filtered)
+    if args.amalgamate:
```
gonuke, line 453: I'll note that this change from `args().` to `args.` would be a great thing to have been a standalone PR - not asking for you to change anything now.

gonuke (approved):

tools/ALARAJOYWrapper/preprocess_fendl3.py
eitan-weinstein: Fair enough. Thanks for letting it slide through here.

PR #287, opened 2026-06-25
Introducing Comparative Cross-Section Plotting Module

### LINKED ISSUE #286: Create Groupwise XS Plotting Function
Within the suite of tools for ALARAJOY preprocessing, build a plotting module to make use of the `matplotlib.pyplot.step()` function for particular spectral data and allow for comparison with continuous TENDL data, outside sources.

eitan-weinstein:
Follows #285. Closes #286.

This PR creates a new module `xs_plotting.py` that can be used within or independently of the core preprocessor to produce stair plots of groupwise cross-sections alongside continuous-energy TENDL cross-sections. 

I've been putting this together following the discussion in the 6/22 meeting about the needs for some plotting tool for ALARAJOY-produced groupwise cross-sections. Given @gonuke's comment on #286 from this morning, these changes may be more expansive than desired for a single PR, but I wanted to open it still as a framing point for coming work on plotting tools.

### DIFF
--- a/src/DataLib/ALARAJOY.cpp
+++ b/src/DataLib/ALARAJOY.cpp
@@ -33,14 +33,20 @@ void ALARAJOYLib::loadDSVData()
     currentParent = -1;
     DSVRow row;
 
-    // Read header containing energy group number
-    inTrans >> nGroups;
+    // Read header containing energy group number in first entry
+    std::string headerLine;
+    std::getline(inTrans, headerLine);
+    std::istringstream headerStream(headerLine);
+    headerStream >> nGroups;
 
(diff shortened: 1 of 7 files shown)


gonuke:
Definitely a lot more here than I would have started with.  While we may ultimately want all of this, this implementation is far more monolithic than we like to see.  Ideally, a specialized plotting function is relatively lightweight, and focuses on selecting the appropriate formatting for a particular set of data, but is agnostic to where the data comes from.  Such a plotting function could then be used to produce multiple plots across lots of data sets, but it would be called as a stand alone function from another place that is determining how to assemble those datasets.  `plot_cross_sections` definitely conflates many purposes.

PR #288, opened 2026-06-25
Cross-Section Plotting -- Basic Function Only

### LINKED ISSUE #286: Create Groupwise XS Plotting Function

eitan-weinstein:
Closes #286.

This PR is working backwards from the over-expansive version in #287, which tackles too many issues at once by simply introducing a cross-section plotting function that is capable of comparatively plotting different groupwise and continuous cross-section vs energy plots together for a given nuclide/reaction.

### DIFF
--- a/tools/ALARAJOYWrapper/xs_plotting.py
+++ b/tools/ALARAJOYWrapper/xs_plotting.py
@@ -72,6 +72,5 @@ def plot_single_nuc_rxn_xs(
     ax.set_title(title)
     ax.grid()
     ax.legend()
-    print(ax.get_title())
 
     return ax
--- a/tools/ALARAJOYWrapper/xs_plotting.py
+++ b/tools/ALARAJOYWrapper/xs_plotting.py
@@ -1,5 +1,5 @@
 def plot_single_nuc_rxn_xs(
-    ax, element, A, emitted, continuous_dict={}, groupwise_dict={},
+    ax, element, A, emitted, continuous_dict={}, groupwise_dict={}
 ):
     """
     Create a plot for a singular nuclide/reaction's cross-sections vs. energy.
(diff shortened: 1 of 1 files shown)


eitan-weinstein:
Currently, I've only included the function definition to keep this PR small, but if you think it's a good idea @gonuke, I can also build in its implementation within `preprocess_fendl3.py`. Otherwise, I can save that for a future PR.

eitan-weinstein:
I also realize that as it stands with just the function definition, this PR is not directly dependent on #285 as far as the code itself goes, but is nevertheless the functional follow-up to the expansion beyond a restrictive, single group structure preprocessor.

gonuke:
I merged #285 but this still includes some of those changes... I am not really sure why, but let's try having you merge `main` back into this branch to see what happens (i.e. NOT a rebase...)

tools/ALARAJOYWrapper/xs_plotting.py
```diff
@@ -0,0 +1,70 @@
...
+        Can be used to plot continuous TENDL data (not processed by ALARAJOY),
+        alongside an arbitrary number of groupwise cross-sections according to
+        the group structure in which they were converted. Groupwise and
+        continuous data can be plotted individually if only one type is
+        provided.
+
+    Arguments:
+        ax (matplotlib.axes._axes.Axes): Matplotlib Axes object of the plot
+            being constructed.
+        single_rxn_dict (dict): Nested dictionary keyed at the highest level
```
gonuke: I think it makes sense to separate this into two variables, one for the multigroup data and one for the continuous data, each can default to `None` and the code can adapt accordingly.

gonuke (changes requested):
Just one request here and then we'll see how it is streamlined following a rebase/merge.

gonuke (approved):

PR #289, opened 2026-06-29
Cross-Section Plotting -- Continuous Data Extraction

eitan-weinstein:
Follows #288.

This PR introduces functionality to extract cross-sections and energies from a continuous-energy TENDL file and stored in a data structure of the form `{'xs' : tendl_xs, 'energies' : tendl_energies}` for the `continuous_dict` parameter in `xs_plotting.plot_single_nuc_rxn_xs()`.

### DIFF
--- a/tools/ALARAJOYWrapper/xs_plotting.py
+++ b/tools/ALARAJOYWrapper/xs_plotting.py
@@ -1,3 +1,4 @@
+import re
 import tendl_processing as tp
 
 def flagged_num_to_int(num):
@@ -13,7 +14,14 @@ def flagged_num_to_int(num):
         integer form.
     """
 
-    return int(''.join(char for char in str(num) if char.isdigit()))
+    re_match = re.match(r'^\d+', str(num))
+    if not re_match:
+        raise ValueError(
+            f'Invalid flagged number {num}. Must be formatted with numeric ' \
+            'characters before non-numeric characters.'
+        )
+    
+    return int(re_match.group())
 
 def extract_continuous_data(tendl_path, MT):
     """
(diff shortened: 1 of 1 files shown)


gonuke:
Note that this PR is entirely independent of #288, even if it may be motivated by it. This is one of the advantages of breaking down problems into smaller pieces with clear separation of concerns. This PR doesn't need to be a branch from #288,  it could be a branch from `main` because it doesn't matter why you want to extract continuous data or how you're going to use it. It may be that your immediate plan is to use it for plotting, but it's a generic capability that may have multiple use cases.

gonuke:
In any case, it now needs a merge/rebase following #288

tools/ALARAJOYWrapper/xs_plotting.py
```diff
@@ -1,3 +1,57 @@
...
+        base integer form.
+
+    num (int or str): Numerical value, which may be in string form with
+        additional tags such as 'm' or '*'.
+
+    num_int (int): Numerical value stripped of any non-numerical characters in
+        integer form.
+    """
+
+    return int(''.join(char for char in str(num) if char.isdigit()))
```
gonuke: We should probably look into regular expressions for things like this
```
match = re.match(r'^\d+', str(num))
if match:
   return int(match.group()
else:
   return 0
```

gonuke: what does this return if `num` doesn't start with digits?

gonuke (changes requested):
Just some questions about the integer matching

tools/ALARAJOYWrapper/xs_plotting.py
eitan-weinstein: It probably should raise a `ValueError` -- I'll work that into your regular expression suggestion.

gonuke (approved):
LGTM

PR #290, opened 2026-06-29
Cross-Section Plotting -- Setting Image Filepaths According to Hierarchical Directory Structure

eitan-weinstein:
Follows #288, independent of #289.

This PR introduces a concise function to recreate the file saving structure originally proposed in #287, but in a much more concise and modular way. This still employs the same hierarchical directory structure of: `TENDL_DATA_SOURCE->ELEMENT->NUCLIDE`, with each reaction plot having its own unique filepath within the nuclide-level directory (with the possibility of multiple versions of a single reaction for different combinations of group structure comparisons).

### DIFF
--- a/tools/ALARAJOYWrapper/xs_plotting.py
+++ b/tools/ALARAJOYWrapper/xs_plotting.py
@@ -115,7 +115,7 @@ def set_plot_save_path(
 
 def plot_single_nuc_rxn_xs(
     ax, element, A, emitted, tendl_dir,
-    continuous_dict={}, groupwise_dict={}, img_ext='png'
+    continuous_dict={}, groupwise_dict={}, saving=True, img_ext='png'
 ):
     """
     Create and save a plot for a singular nuclide/reaction's cross-sections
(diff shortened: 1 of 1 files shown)


tools/ALARAJOYWrapper/xs_plotting.py
```diff
@@ -1,13 +1,68 @@
...
+    nuc = f'{element}{A}'
+    nuc_dir = Path(f'{tendl_dir}_plots') / element / nuc
+    nuc_dir.mkdir(parents=True, exist_ok=True)
+
+    return nuc_dir / f'{nuc}_(n,{emitted})_{"_".join(group_names)}.{img_ext}'
+
 def plot_single_nuc_rxn_xs(
-    ax, element, A, emitted, continuous_dict={}, groupwise_dict={}
+    ax, element, A, emitted, tendl_dir,
+    continuous_dict={}, groupwise_dict={}, saving=True, img_ext='png'
```
gonuke: Instead of passing this extra data to generate a file name, why not generate the filename externally and just pass that
```suggestion
    continuous_dict={}, groupwise_dict={}, save_filename=None
```

gonuke: I guess this makes it default to no plot, but that may be better anyway

gonuke: Or... since this method returns the axes, can you just save it after it returns 

(All of these comments being about separation of concerns)

gonuke (changes requested):

tools/ALARAJOYWrapper/xs_plotting.py
eitan-weinstein: Sure, I think for the separation of concerns, it makes sense to just exclude all saving operations from the plot construction entirely.

gonuke (approved):
LGTM

gonuke:
Looks like there are merge conflicts

eitan-weinstein:
Just did a rebase. Merge conflicts should be resolved now.

gonuke (approved):

PR #291, opened 2026-06-30
Cross-Section Plotting -- Accessing Group Structure Bounds

eitan-weinstein:
This PR is motivated by the need to procure groupwise energy bounds for cross-section plotting. Two functions are introduced to `njoy_tools.py` to accomplish this:

1) `load_external_group_struct()`. This function takes the already-existing conditional block for the reference file and OpenMC options from `set_group_structure()` and puts them into their own function, which is then called in the same location within `set_group_structure()`. The purpose of this is to be able to use `load_non_njoy_group_struct()` to collect the energy bounds needed to produce a stair plot for groupwise data, without the reformatting of the group bounds to the joined `egn` string that is output from `set_group_structure()`.

2) `copy_energy_groups_from_njoy_output()`. This function reads the group bounds from an NJOY/GROUPR output file and copies them to a new file with the name of the group structure. This will create reference files suitable for reading in `load_external_group_struct()`, which is necessary in the case that the group structure was otherwise one of the built-in NJOY options.

### DIFF
--- a/tools/ALARAJOYWrapper/njoy_tools.py
+++ b/tools/ALARAJOYWrapper/njoy_tools.py
@@ -601,7 +601,9 @@ def run_njoy(element, A, matb, file_capture):
             
     return fileinfo['save'], result.stderr
 
-def copy_energy_groups_from_njoy_output(group_name, nGroups):
+def copy_energy_groups_from_njoy_output(
+    group_name, nGroups, groupr_output='output'
+):
     """
     Copy the energy bounds printed out to an NJOY output file from a GROUPR
         run to a new file in ascending order.
@@ -611,12 +613,15 @@ def copy_energy_groups_from_njoy_output(group_name, nGroups):
             GROUPR converted continuous energy cross-sections to groupwise
             cross-sections.
         nGroups (int): Number of groups in the given group structure.
+        groupr_output (str or pathlib._local.PosixPath, optional): Option to
+            set filepath for NJOY/GROUPR output from which to read in energy
+            group bounds.
 
     Returns:
         None
     """
 
-    with open('output', 'r') as f:
+    with open(groupr_output, 'r') as f:
         lines = f.readlines()
 
     last_group_line = next(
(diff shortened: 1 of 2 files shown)


tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -556,6 +601,43 @@ def run_njoy(element, A, matb, file_capture):
             
     return fileinfo['save'], result.stderr
 
+def copy_energy_groups_from_njoy_output(
```
gonuke, line 604: This seems a little circular... you're planning to read group info from NJOY output so that you can load it into NJOY?

gonuke (changes requested):

tools/ALARAJOYWrapper/njoy_tools.py
eitan-weinstein: While that particular use-case is possible and somewhat circular, the intended purpose of this function is to store the energy group bounds from an NJOY run into a filetype consistent with the formatting that would otherwise be used to read in arbitrary bounds for running NJOY. That way, for the various uses we might have for reading these groups, we can re-use the function `load_external_group_struct()`. 

Looking at #292 or #295, you can see this intended purpose in-action, where groupwise energy bounds are loaded for plotting purposes. If the group structure is not otherwise contained in `openmc.mgxs.GROUP_STRUCTURES`, there would need to be a reference file containing the bounds. If the groupwise data was produced from one of the built-in NJOY group structures, however, the only way to get those energy bounds is from the output itself (or probably by digging in deep to the NJOY source code to copy the values from wherever they may be stored, but this seems far less preferable to produce the same kind of reference file anyways).

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -556,6 +601,43 @@ def run_njoy(element, A, matb, file_capture):
...
+        None
+    """
+
+    with open(groupr_output, 'r') as f:
+        lines = f.readlines()
+
+    last_group_line = next(
+        i - 1 for i, line in enumerate(lines)
+        if ' weight function......' in line
+    )
```
gonuke, line 630: This block is a little obtuse, we should discuss it in hacker within...

gonuke (approved):
LGTM

PR #292, opened 2026-07-01
Cross-Section Plotting -- Including Plotting Option/Implementation to Preprocessing Pipeline

eitan-weinstein:
Follows #291, #290, #289.

This PR introduces a new argument `--xs_plotting` or `-p` to the ALARAJOY data preprocessing pipeline that, when applied, utilizes tools from `xs_plotting.py` to create and save plots for each reaction contained in the converted groupwise data.

In order to capture all reactions without introducing another slow multi-iteration through all parents, daughters, and MTs, I modified the function `preprocess_fendl3.py.write_dsv()` to now be responsible for all data write-outs and thus renamed to `store_results()`.


gonuke:
this needs a rebase following recent merges of other PRs

gonuke (approved):

PR #293, opened 2026-07-01
Cross-Section Plotting -- Including Group Name/MTs in DSV

eitan-weinstein:
This PR makes a minor modification to the canonical format an ALARAJOY DSV to now include the name of the group structure according to which the groupwise data was converted in the first line after `nGroups` and for each reaction row to include the MT value corresponding to said reaction. Accordingly, `src/DataLib/ALARAJOY.cpp` is updated to handle this new format, without changing the data that it passes on to the rest of the ALARA binary library construction pipeline.

The motivation for this PR is that in order to create a standalone script for `xs_plotting` that will allow users to supply different DSV files as input, these additional pieces of data will be useful. Particularly being able to read in the group name from a DSV file is helpful in clarifying beyond just the group number count, should there be any group structures that share the same number of groups but which are not the same.

### DIFF
--- a/src/DataLib/ALARAJOY.cpp
+++ b/src/DataLib/ALARAJOY.cpp
@@ -33,14 +33,20 @@ void ALARAJOYLib::loadDSVData()
     currentParent = -1;
     DSVRow row;
 
-    // Read header containing energy group number
-    inTrans >> nGroups;
+    // Read header containing energy group number in first entry
+    std::string headerLine;
+    std::getline(inTrans, headerLine);
+    std::istringstream headerStream(headerLine);
+    headerStream >> nGroups;
 
     // Extract Parent KZA until EOF at pKZA == -1
     while ((inTrans >> row.parentKZA) && row.parentKZA != -1)
     {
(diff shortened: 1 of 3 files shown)


src/DataLib/ALARAJOY.cpp
```diff
@@ -33,14 +33,20 @@ void ALARAJOYLib::loadDSVData()
     currentParent = -1;
     DSVRow row;
 
-    // Read header containing energy group number
-    inTrans >> nGroups;
+    // Read header containing energy group number in first entry
+    std::string headerLine;
+    std::getline(inTrans, headerLine);
+    std::istringstream headerStream(headerLine);
+    headerStream >> nGroups;
```
gonuke:
```suggestion
    // Read header containing energy group number in first entry
    std::string groupName;
    inTrans >> nGroups >> groupName;
```

gonuke (changes requested):
One little change where you added more info to header

gonuke (approved):
LGTM

PR #294, opened 2026-07-01
Cross-Section Plotting -- Groupwise Data Extraction from DSV

eitan-weinstein:
Follows #292.

Reliant on outputs from #293, but code is technically independent.

This PR introduces a new function in `xs_plotting.py` to produce a dictionary of groupwise data keyed by group structure name to be used as an input for `plot_single_nuc_rxn_xs()`. The function takes in a list of DSV files to read, as well as nuclide/reaction identifying information. It parses the group structure name from the DSV file (added in #293) and then goes through line-by-line until the provided reaction is found at which point it is added to the dictionary.

### DIFF
--- a/tools/ALARAJOYWrapper/xs_plotting.py
+++ b/tools/ALARAJOYWrapper/xs_plotting.py
@@ -220,7 +220,6 @@ def ensure_emission_specificity(emitted, dKZA):
 
     return emitted
 
-
 def flagged_num_to_int(num):
     """
     Convert numerical values that may be in string form containing additional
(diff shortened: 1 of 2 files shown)


tools/ALARAJOYWrapper/xs_plotting.py
```diff
@@ -26,6 +28,32 @@ def flagged_num_to_int(num):
     
     return int(re_match.group())
 
+def ensure_emission_specificity(emitted, dKZA):
```
gonuke, line 37: Is this a duplicate now?

gonuke (changes requested):
seems like a duplicate function has popped back in

tools/ALARAJOYWrapper/xs_plotting.py
eitan-weinstein: No, I just checked -- it only appears once

gonuke: I see - it's just being moved...

gonuke:
I'll take another look once we address merge conflicts

PR #295, opened 2026-07-01
Cross-Section Plotting -- Making xs_plotting.py work as a standalone script

eitan-weinstein:
Follows #294.

This PR pulls together features that have been developed in #288 - #294 allowing `xs_plotting.py` to be called as a standalone script to plot groupwise data from any number of DSV files for any number of user-selected nuclides and reactions. This follows some of the intended output from #287, but in a much more streamlined and modular way (further evidence of the value in breaking these into smaller PRs...). I have kept the YAML input structure from #287, but the iterative implementation is now much more efficient and does not rely on producing empty sub-directories only to be deleted post-facto.


gonuke:
This may need another rebase now?

gonuke (approved):
This is probably fine if it works for you...

PR #296, opened 2026-07-02
Remove ta-180 from elemental composition

anu1217:
PRs #73 and #74 added ta-180 into the elemental composition of tantalum. This was motivated by an attempt to match the input nuclides accepted by OpenMC at the time. However, OpenMC has since removed ta-180 from the material composition of the element, and defaults to using ta-181 in their definition of the natural abundance. 

This is one of the few cases where the excited state (ta-180m1) is more stable than the ground state. While this might motivate a switch to using the metastable state in place of the ground state, some older cross-section libraries do not contain data on ta-180m1. As a result, OpenMC defines ta-181 as being 100% of the natural abundance, and I think ALARA should follow suit.

(To my knowledge, ALARA doesn't yet have the capability to accept metastable states in the initial material composition.)


gonuke (approved):
LGTM

PR #297, opened 2026-07-08
Fixing MT variable type in ALARAJOY DSV parser

eitan-weinstein:
This PR fixes a small error introduced in #293 in which the temporary `MT` variable was set as an integer. This was problematic because MT values that correspond to excitations reactions will have a `'*'` appended to them and therefore cannot be read in as integers. Given that the MT value in the `ALARAJOY::loadDSVData()` is only needed for temporary storage, changing this variable type to a string resolves this issue without need for further editing of the MT value post-facto.


gonuke (approved):
I guess this matters because you modify MT numbers so that they aren't always integers?

PR #299, opened 2026-07-08
Fixing `flagged_num_to_int()` negative number handling

eitan-weinstein:
This PR fixes a small bug in `xs_plotting.flagged_num_to_int()` regular expression handling that excluded negative numbers to be treated as readable integers. This is problematic because the data preprocessor can output MT values of -1 for the `special_MT` handling in `tendl_processing.iterate_MTs()`.


gonuke (approved):
seems straightforward

PR #302, opened 2026-07-15
Saving Isomer '*' Flag for `flagged_num_to_int()` plotting function

eitan-weinstein:
This PR modifies the function `xs_plotting.flagged_num_to_int()` to additionally return a string of all of the `'*'` isomer flags in the value passed through it. For example, the input `'4*'` would return two values: `(4, '*')`. Likewise, an input without any `'*'` values will return an empty string for the `isomer_flag` return value.

The motivation behind this is to be able to retain the distinction between MT values in which there may be present anywhere from 0-2 `'*'` flags for a given target nuclide. Without this update, all of these distinct reactions would be improperly characterized by only the star-less MT.


gonuke:
Is this now incorporated into #305? Do we still need this PR?

Issue #303, opened 2026-07-15
Create continuous cross-section extraction method for MF9/10 state-specific reactions


eitan-weinstein:
Currently, `ALARAJOYWrapper/xs_plotting.py` extracts continuous cross-sections according to MT values that have been reformatted as integers, but may have otherwise included isomeric flags (i.e. `'*'` for a reaction specifying the first excited state of a daughter nuclide). Accordingly, continuous cross-sections are only extracted from MF 3 of the unmodified TENDL files, which necessarily excludes state-specific cross-sections or multiplicities that may be contained in MF9/10 (and from which the flagged-MT cross-sections were extracted during the ALARAJOY processing). 

In order to avoid unnecessarily suppressing/overwriting these reactions when producing cross-section plots, `xs_plotting.extract_continuous_data()` should be updated to handle MT values containing isomeric flags by first checking to see if the MF10/MT combination exists, and then secondarily looking for MF9/MT if the former does not. MF10 contains cross-section values, so they can be extracted directly, but MF9 contains energy-dependent reaction multiplicities, which would need to be multiplied by the cumulative cross-sections from MF3 to properly capture the state-specific cross-section values.

Currently, this method uses `tendl_processing.create_endf_file_obj()`, which itself is built around `ENDFtk`. `ENDFtk` does not contain capabilities to parse the multiplicity tables from MF9 beyond just reading in the section from the file. Alternatively, `endf-parserpy` does have this capability, but would necessitate the re-incorporation of another domain-specific software package for only a narrow application.

Regardless of the ENDF parsing software used, I am thinking that the best way to calculate the cross-sections from MF9 will be to create energy-keyed dictionaries from both MF3 and MF9, valued by the cross-section or multiplicity respectively. Each cross-section can then be multiplied for each matching energy, with all other values being 0 (this is necessary because there can be different energies /numbers of energies in MF3/9 for a given reaction).

eitan-weinstein:
Given the option of using `endf-parserpy` I am also considering migrating `ALARAJOYWrapper` away from `ENDFtk` in general in favor of `endf-parserpy`, and this may be a good impetus for that...

PR #304, opened 2026-07-16
Migrating from ENDFtk to endf_parserpy for ENDF parsing

eitan-weinstein:
This PR replaces all previous dependence on the NJOY `ENDFtk` software package in favor of the IAEA-NDS `endf_parserpy` package, while retaining all of the same core functionality in the `tendl_processing` module. The motivation for this change is twofold:

1) As laid out in #303, `ENDFtk` is unable to process MF9 data beyond just reading in the section text into a string, whereas `endf_parserpy` has built-in parsing methods capable of specifically extracting the energies and multiplicities from MF9 sections.  While this is not necessary for the main ALARAJOYWrapper preprocessing methods, given that the same functionality that have already been built with `ENDFtk` can be accomplished with minimal disruption using `endf_parserpy`, migrating entirely to the latter reduces unnecessary domain-specific dependencies.

2) `endf_parserpy` stores parsed ENDF data in a nested dictionary of the form:
     ```
    MF:
        MT:
           section_data
    ```
    I find this data structure to be more transparent and straightforward to work with than `ENDFtk`'s object-oriented structures. Because the MT values are the keys for the sub-dictionaries, this migration actually allows for a simplification of certain methods in `tendl_processing.py`, such as `determine_all_excitations()`, which no longer relies on a separate positional search for the `LFS` values, but instead reads them from already-stored values in the nested dictionary. Furthermore, this eliminates entirely the need for `extract_endf_specs()`, as the MTs can be simply read from the keys of any given MF-dictionary, without any convoluted conditionals.

### DIFF
--- a/tools/ALARAJOYWrapper/README.md
+++ b/tools/ALARAJOYWrapper/README.md
@@ -21,7 +21,7 @@ This preprocessor uses [NJOY 2016](https://github.com/njoy/NJOY2016) Nuclear Dat
   * [NumPy](https://numpy.org/install/)
   * [Pandas](https://pandas.pydata.org/docs/getting_started/install.html)
 - Domain-specific packages
-  * [ENDFtk](https://github.com/njoy/ENDFtk)
+  * [Endf-parserpy](https://github.com/IAEA-NDS/endf-parserpy)
   * [NJOY 2016](https://github.com/njoy/NJOY2016)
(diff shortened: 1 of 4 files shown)


gonuke (approved):
Looks OK to me. I don't know enough about the details to really judge if it's correct, but presume you'd see it in the results.

PR #307, opened 2026-07-24
Including ratio plotting functionality to `xs_plotting`

eitan-weinstein:
Functionally follows #314 and #317, but only in DSV output formatting, not any direct code dependency.

Similar to the ratio plotting capabilities in `alara_output_processing.alara_output_plotting.py`, this PR introduces new functionality to `xs_plotting.py` to allow for the production of plots showing the ratios of groupwise cross-sections across different group structures. Given the inherent differences in energy group bounds/number of groups between different group structures, this PR includes a function `compute_groupwise_xs_ratios()` that constructs a union grid of the energies between both group structures and pulls the appropriate cross-sections at those energies from the existing groupwise values. Unlike the interpolation approach in #305, this is merely pulling existing cross-section values that fall within re-adjusted energy bins so that like-sized and matching-index arrays between two group structures of different sizes can be appropriately compared. These are then divided using `np.divide()`.

This functionality will be useful in diagnosing more precisely the effects of choice of group structure has on any given activation cross-sections.

### DIFF
--- a/tools/ALARAJOYWrapper/xs_plotting.py
+++ b/tools/ALARAJOYWrapper/xs_plotting.py
@@ -134,20 +134,12 @@ def extract_groupwise_data_from_DSV(dsv_list, KZA, MT):
 
     groupwise_dict = {}
     emitted = ''
-    reference_path = ''
-    reference_group = ''
 
     for dsv in dsv_list:
-        if 'reference' in dsv.lower():
-            reference_path = dsv.split('reference')[0].strip('(').strip()
-            dsv = reference_path
-
         with open(dsv, 'r') as f:
             dsv_lines = f.readlines()
 
         group_name = dsv_lines[0].split()[-1]
-        if dsv == reference_path:
-            reference_group = group_name
 
         _, energy_bounds = njt.load_external_group_struct(group_name)
 
@@ -163,7 +155,7 @@ def extract_groupwise_data_from_DSV(dsv_list, KZA, MT):
                 }
                 break
 
-    return groupwise_dict, emitted, reference_group
+    return groupwise_dict, emitted
 
 def set_plot_save_path(
     element, A, emitted, tendl_dir, group_names,
@@ -364,23 +356,17 @@ def collect_all_stair_colors(ax):
 
     return color_dict
 
-def compute_groupwise_xs_ratios(groupwise_dict, reference_group=''):
+def compute_groupwise_xs_ratios(groupwise_dict):
     """
     Calculate the ratios of groupwise cross-sections for each group in
-        `groupwise_dict` and a single reference series. For each group
-        pairing, adjust the energies and cross-sections to be on a merged
-        grid that accounts for potentially distinct bin boundaries between
-        energy groups.
+        `groupwise_dict` that shares a group structure as a single reference
+        series. `ValueError` is raised if no other groupwise data matches the
+        group structure of the reference group structure.
 
     Arguments:
         groupwise_dict (dict): Nested dictionary keyed at the highest level by
             the name of the group structure according to which an array of
             cross-sections were processed.
-        reference_group (str, optional): Group name of the reference group
-            structure against which to divide other group structures' cross-
-            sections. If not supplied, then the first group name key in
-            `groupwise_dict` will be set by default as the reference group.
-            (Defaults to '')
 
     Returns:
         ratio_dict (dict): Nested dictionary keyed at the highest level by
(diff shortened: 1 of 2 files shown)


tools/ALARAJOYWrapper/xs_plotting.py
```diff
@@ -125,33 +126,48 @@ def extract_groupwise_data_from_DSV(dsv_list, KZA, MT):
             cross-sections were processed.
         emitted (str): Particle(s) emitted from the nuclear reaction
             corresponding to the MT number provided.
+        reference_group (str): Group name of the reference group against which
```
gonuke, line 129: Separation of concerns: can't we just have a method that only does this and we can avoid encoding "(reference)" into the string?

tools/ALARAJOYWrapper/xs_plotting.py
```diff
@@ -190,27 +209,73 @@ def set_plot_save_path(
...
+                VITAMIN-J-175_CCFE-709_ratios.png
     """
 
     if isinstance(group_names, str):
         group_names = [group_names]
 
     nuc = f'{element}{A}'
     nuc_dir = Path(f'{tendl_dir}_plots') / element / nuc
+    if ratio_plotting:
+        nuc_dir /= 'ratio_plots'
```
gonuke, line 227: If you do this, you can make the path creation below simpler 

```suggestion
    ratio_suffix = ""
    if ratio_plotting:
        nuc_dir /= 'ratio_plots'
        ratio_suffix = "_ratios"
```

tools/ALARAJOYWrapper/xs_plotting.py
```diff
@@ -272,15 +334,203 @@ def plot_single_nuc_rxn_xs(
...
+            section data divided by the reference group's (according to the
+            re-computed bins to appropriately match cross-sections and
+            energies between the two group structures).
+        reference_group (str): Group name of the reference group structure.
+            Will be the same as `reference_group` from Arguments if one is
+            supplied, otherwise, will be the group name of the first key in
+            `groupwise_dict`.
+    """
+
+    def _re_bin_xs(bin_edges, xs, midpoints):
```
gonuke, line 400: We should discuss this approach.  I wonder if you can put some diagrams together with all of the different cases you expect to see.  Rebinning group x-sections is a fraught topic...

gonuke (changes requested):

tools/ALARAJOYWrapper/xs_plotting.py
eitan-weinstein: Given our discussion last week and my introduction of the proposed changes in #314 that would allow us to compare directly against FISPACT-II's nuclear data, I'm going to abandon the re-binning approach. Instead, I'm going to necessitate that all ratio-plotted comparisons must have the same group bounds/nGroups so that cross-section division is unambiguously non-problematic.

tools/ALARAJOYWrapper/xs_plotting.py
eitan-weinstein: We could even simplify this further by not including any explicit `reference` tag, but rather just relying on ordering in the DSV section of the YAML. in `compute_groupwise_xs_ratios()`, the default option for the case when no reference is explicitly provided is to just take the first key in `groupwise_dict` with the command:

```
reference_group = reference_group or next(iter(groupwise_dict))
```

If we just specify in the documentation that the way to set the reference group is by placing its DSV path first amongst the DSVs in the YAML, then this can be accomplished much simpler and with cleaner inputs.

gonuke: So, I think we can delete this entry in the comments now, right?

tools/ALARAJOYWrapper/xs_plotting.py
```diff
@@ -125,33 +126,47 @@ def extract_groupwise_data_from_DSV(dsv_list, KZA, MT):
...
+            string if none of the provided DSV paths in `dsv_list` contain a
+            '(reference)' tag in the string (i.e. example.dsv (reference)).
     """
 
     groupwise_dict = {}
     emitted = ''
+
     for dsv in dsv_list:
         with open(dsv, 'r') as f:
             dsv_lines = f.readlines()
```
gonuke: Not critical, but something to think about: this could be written in a way that reads one line at a time...

tools/ALARAJOYWrapper/xs_plotting.py
gonuke: Re-upping this comment

tools/ALARAJOYWrapper/xs_plotting.py
```diff
@@ -272,15 +337,164 @@ def plot_single_nuc_rxn_xs(
...
+            other group names are implicitly refering to that group's cross-
+            section data divided by the reference group's. All data must be of
+            the same group structure.
+        reference_group (str): Group name of the reference group structure.
+            Will be the same as `reference_group` from Arguments if one is
+            supplied, otherwise, will be the group name of the first key in
+            `groupwise_dict`.
+    """
+
+    reference_group = next(iter(groupwise_dict))
```
gonuke: Why not make this an input to the method, requiring the user to name their reference, rather than inferring it?

The user could infer it in the same way from outside the method before calling this method, if they like.

gonuke (changes requested):
A few additional comments

tools/ALARAJOYWrapper/xs_plotting.py
eitan-weinstein: Sure, I may as well change that over. Shouldn't be too big of a task, saves some memory.

eitan-weinstein:
Putting on pause until after #333 so that metadata parsed from DSV header can be used for series/plot identification.

PR #310, opened 2026-07-29
Migrating ENDF interfacing from `endf_parserpy` to `openmc.data`

eitan-weinstein:
Follows #305.

Even though we only recently made the switch from `ENDFtk` to `endf_parserpy` for our internal ENDF/PENDF interfacing, taking advantage of OpenMC's NJOY-compatible interpolation functionality pushed me to further investigate using `openmc.data` in general, beyond just the use-case of continuous energy cross-section interpolation within `xs_plotting`.

My main concern going into this, as I mentioned in the thread under #305, was whether `openmc.data` would be able to parse PENDF files. However, after building these new updates and testing them against the results with the previous `endf_parserpy` dependency, I found them to agree and that concern turned out to be a non-issue. Moreover, a nice add-on benefit that I've found from this migration is that OpenMC's parsing seems to be _much_ faster than `endf_parserpy`. Not that it matters so much in the grand scheme of things, because the main contributing factor to the speed of the preprocessor is actually running NJOY itself, but it's still nice to improve the operating speed where it can be done.


gonuke:
I think this now has merge conflicts to be resolved from #305?

gonuke (approved):
Looks good to me - thanks @eitan-weinstein

PR #314, opened 2026-08-03
Building in PREPRO-processed activation data (as formatted for FISPACT-II)

### LINKED ISSUE #311: Build parsing capabilities for FISPACT-formatted GENDF files
In order to further the validation efforts for ALARAJOY, it will be valuable to have functionality to read in nuclear data from the TENDL GENDF data readable by FISPACT-II, which is distributed at https://git.oecd-nea.org/fispact/nuclear_data/. Unlike ALARAJOY, which preprocesses nuclear data using NJOY, with the specific groupwise conversion being done with the `GROUPR` module, FISPACT-II's data is prepared with PREPRO, with its `GROUPIE` module for groupwise conversion. As a result, the specific formatting of the respective GENDF files are not identical, requiring modified parsing methods to accommodate PREPRO GENDF files.

Accomplishing this will be primarily valuable for two purposes:
1) FISPACT-II/PREPRO groupwise cross-sections (or reaction rates) can be compared directly against ALARAJOY cross-sections to assess fundamental similarity in grouping outcomes, independently of any activation simulations.
2) With minimal further modification, a directory of FISPACT-II/PREPRO GENDF files could simply be provided to ALARAJOYWrapper and have its cross-section data read and transcribed to an ALARAJOY DSV, allowing ALARA to be run using the _exact_ same data as FISPACT-II. Note, this goal is slightly separate from the ALARAJOY validation efforts per se, because this would not be testing the validity of the ALARAJOYWrapper preprocessor itself, but rather just borrowing limited functionality for data reading, organizing, and writing, without performing any calculations. Nevertheless, this could be valuable in general for users wishing to do direct FISPACT-II/ALARA comparisons with identically preprocessed data. Inasmuch as the ALARAJOY work done so far has aimed at allowing TENDL data to be processed as similarly as possible to FISPACT-II's, fundamentally NJOY and PREPRO are handling the data processing differently and may produce somewhat different results in various cases.

eitan-weinstein:
Closes #311.

This PR introduces the capability for `tendl_processing.py` to parse groupwise nuclear data produced by PREPRO (via its `GROUPIE` module) and incorporate it into the ALARAJOY workflow. This is motivated by the desire to compare input groupwise cross-section data between ALARA and FISPACT-II prior to any activation simulations. Unlike the ALARAJOY workflow, which allows users to create their own ALARA binary libraries from TENDL data, FISPACT-II's nuclear data is distributed in it's preprocessed, ready-to-go format through [FISPACT-II's NEA GitLab nuclear data repository](https://git.oecd-nea.org/fispact/nuclear_data). 

There are two main formatting differences between NJOY/GROUPR-produced GENDF files and those produced by PREPRO/GROUPIE:
1) For cross-section data, `GROUPR`-formatted GENDF files will only contain MF3. To produce excitation pathway-specific cross-sections, MF9/10 data can be highlighted during input (see #229), but will ultimately be written out in a subsection format _within_ MF3 to the final GENDF output. `GROUPIE`, on the other hand, preserves MF10 in its GENDF formatting, meaning MF3 will not contain any subsections, just the cumulative reaction cross sections for all pathways (if multiple exist). MF10 contains the pathway data itself.
2) Actual formatting of each TAB1 data table differs between `GROUPR` and `GROUPIE`, requiring different parsing approaches.

To accommodate these differences, I restructured the main data structure in which to store parsed GENDF data to now being a dictionary keyed by `MF` and valued by a sub-dictionary with keys `'MTs'` and '`non_zero_xs'` (adapting the nomenclature from my original architecture). The definition of this dictionary is as such:

```
{
      MF : {
          'MTs' : set(),
          'non_zero_xs' : (
              defaultdict(dict) if MF == 10 else defaultdict(list)
          )
      }
      for MF in self.MFs
 }
```

The special case of `MF == 10` for `non_zero_xs` allows for the GROUPIE MF10 subsections to be stored with each LFS as the key to a similar list-of-dictionary structure that would otherwise appear at that same level for the standard MF3 case. 

The biggest actual changes that I've made in the code itself, however, is in the creation of a new class `GENDFParser` that stores all of these internal methods relating to the different possibilities for GENDF parsing. As I was building this, I was considering that because there were so many free-floating functions within `tendl_processing.py` that _only_ pertained to use within `extract_gendf_data()`, that it would be most appropriate to organize these all within a class. I know that we had discussed that in general, that it would probably not be worth the time/effort to restructure all of ALARAJOYWrapper to an object-oriented architecture, however, in this limited case, it seemed appropriate to me to conceptually distinguish all of the parsing-specific functions from the rest of the `tendl_processing` module. With all of that said, I recognize that it makes the size of this PR rather large, at least as far as number of lines of code changed is concerned. If you think it would be preferable @gonuke to make a separate preceding PR that just builds out the already existing GENDF parsing functionality to a `GENDFParser` class, without the `GROUPIE` capabilities, I could do work backwards to do so.

### DIFF
--- a/src/DataLib/ALARAJOY.cpp
+++ b/src/DataLib/ALARAJOY.cpp
@@ -35,7 +35,8 @@ void ALARAJOYLib::loadDSVData()
 
     // Read header containing energy group number in first entry
     std::string groupName;
-    inTrans >> nGroups >> groupName;
+    std::string processingCode;
+    inTrans >> nGroups >> groupName >> processingCode;
 
     // Extract Parent KZA until EOF at pKZA == -1
     while ((inTrans >> row.parentKZA) && row.parentKZA != -1)
--- a/tools/ALARAJOYWrapper/README.md
+++ b/tools/ALARAJOYWrapper/README.md
@@ -38,6 +38,10 @@ Many decay data libraries are distributed in repositories containing individual 
   * [TENDL](https://tendl.imperial.ac.uk/)
      - Usable with any TENDL release
      - For FENDL3.2x processing, [TENDL 2017](https://tendl.imperial.ac.uk/tendl_2017/tendl2017.html) is the standard version.
+  * [PREPRO-Processed TENDL Data (FISPACT-II)](https://git.oecd-nea.org/fispact/nuclear_data)
+     - Nuclear data repository for use specifically with FISPACT-II.
+     - TENDL 2017/14 availble.
+     - Convertable to ALARAJOY-readable DSV file by calling `preprocess_fendl3` with `-n` flag.
 
 - Decay Data
   * EAF
@@ -55,7 +59,7 @@ ALARAJOYWrapper is designed to produce a space-delimited DSV containing cross-da
 
 Running ALARAJOYWrapper can be done with one Python command:
 ```
-python preprocess_fendl3.py -f /path/to/fendl3_data_dir/ -d /path/to/decay_library/ decay_library-type -g group_name -a -t -r -p
+python preprocess_fendl3.py -f /path/to/fendl3_data_dir/ -d /path/to/decay_library/ decay_library-type -g group_name -a -t -r -p -n
 ```
 To read in detail about each of these arguments, call this command:
 ```
--- a/tools/ALARAJOYWrapper/preprocess_fendl3.py
+++ b/tools/ALARAJOYWrapper/preprocess_fendl3.py
@@ -70,6 +70,15 @@ def make_argparser():
                 sections from the original TENDL file.
         ''')
     )
+    parser.add_argument(
+        '--nea_prepro', '-n', action='store_true',
+        help=('''
+            Option to convert PREPRO-formatted CCFE-709 groupwise data into
+            an ALARAJOY-formatted DSV. If true, must also provide the path to
+            the directory containing PREPRO .asc GENDF files for the `-f`
+            argument.
+        ''')
+    )
     return parser
 
 def configure_logging(redirect_warnings=False):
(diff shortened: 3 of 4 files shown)


eitan-weinstein:
Something interesting that I am coming across following my updates in #317 and producing comparative cross-section plots between FISPACT-II PREPRO data and ALARAJOY data is a significant mismatch in cross-section positioning for almost all reactions, with the exception of (n,γ) and (n,p) (at least within the limited scope of <sup>54</sup>Fe that I've checked thus far today). For example, there is very clear alignment in the (n,γ) case, with some small discrepancies in the upper resonance range, but not terribly concerning:

image

On the other hand, *all* other reactions for <sup>54</sup>Fe show what appears to be a misalignment that would be resultant from the order reversal of the PREPRO energy groups (rather than cross-sections):

image


At face value, it's hard to decipher the meaning of this. In the current ALARAJOY methodology, cross-section arrays are reversed before they are saved to the output DSV because the order of energy groups in TENDL is opposite to ALARA's expected ordering. If, across the board (including (n,γ), (n,p)), signs of a falsely applied double-reversal were the case, then we'd have a sign that PREPRO-output GENDF files follow the same group structure ordering as we need in ALARA. Given that that's not the case, however, I'm left wondering whether the energy group ordering is inconsistent across PREPRO output data or some other fundamental processing issue in the PREPRO workflow (which I find unlikely) or there is some other hidden bug that may explain how we're getting structurally different results across all reaction types except MT = 102. Having just discovered this, I will continue digging around and seeing if I can better identify the source of this discrepancy.

gonuke:
I'll review after merge conflicts are resolved

PR #317, opened 2026-08-05
Including weight function specification into ALARAJOYWrapper

### LINKED ISSUE #312: Create Arbitrary Weight Function Option for GROUPR
Similarly to the efforts of #285 which opened up ALARAJOYWrapper to take in any arbitrary group structure, the same should be done for GROUPR weighting functions. Not only would this be valuable in giving users more flexibility in their choice of how they want to prepare their data, but it would also allow us to prepare data even more similarly to FISPACT-II's PREPRO-processed groupwise nuclear data.

Related to the aims of #311, while I was more closely inspecting the contents and formatting of FISPACT-II's TENDL-2017 data, I noticed that the listed weight function was `1/E`, rather than the VITAMIN-E piecewise function that is currently employed as the only option for ALARAJOY preprocessing. I found this result to be surprising, given that the stated groupwise processing for [FENDL3 is explicitly weighted by VITAMIN-E](https://www.sciencedirect.com/science/article/pii/S0090375224000012) (Schnabel et al., 2024). Therefore, for the purposes of producing a canonical ALARA library for FENDL3.2x should still be done using VITAMIN-E, but to best compare against FISPACT-II, a `1/E` weighted library should also be produced.

eitan-weinstein:
Closes #312.

This PR enables ALARAJOYWrapper to process groupwise cross-sections with weight functions other than the standard Vitamin-E, that was otherwise hard-coded as a pre-set GROUPR parameter in `njoy_tools`. Primarily, this is accomplished by abstracting the function `njoy_tools.set_group_structure()` to being `njoy_tools.set_modifiable_groupr_parameters()`. The logic for implementing group structure vs weight function specifications is nearly identical, so I've combined both into an agnostic function wherein the respective command line argument and reference dictionaries are supplied to extract either group structure or weight function parameters. In its implementation within `tendl_to_alarajoy.main()`, this looks like:

```
ign, group_name, ngn, egn = njt.set_modifiable_groupr_parameters(
      args.group_structure, njt.NJOY_GROUPS
)
iwt, weight_function, _, _ = njt.set_modifiable_groupr_parameters(
      args.weight_function, njt.NJOY_WEIGHT_FUNCTIONS
)
```

Unlike the group structure handling, I did not include functionality for _arbitrary_ weight function applications, even though NJOY is capable of accommodating them. My main reasoning was that I was building this new functionality specifically for the purpose of having the ability to do `1/E` weighting, which is included among the built-in NJOY weight functions, so that I could have a 1:1 comparison to FISPACT-II's PREPRO-processed TENDL data. While I could conceive  of a user theoretically wanting to create a fully customized weight function themselves (although Vitamin-E is still the standard FENDL weight function), the NJOY implementation is considerably more complex than for arbitrary group structures, which really only require the group bounds. Given that, it did not seem like a worthwhile way to spend my time at this point to build the most expansive arbitrary weight function capabilities, but rather to limit the scope to NJOY's built-in options, which will still allow us to make the necessary comparisons with FISPACT-II.

### DIFF
--- a/tools/ALARAJOYWrapper/njoy_tools.py
+++ b/tools/ALARAJOYWrapper/njoy_tools.py
@@ -86,7 +86,7 @@ groupr_input = Template(Template(
 """
 groupr/
  $NENDF $NPEND $NGOUT1 $NGOUT2/
- $mat_id $ign $IGG $IWT $LORD $NTEMP $NSIGZ $IPRINT_groupr/
+ $mat_id $ign $IGG $iwt $LORD $NTEMP $NSIGZ $IPRINT_groupr/
  $title/
  $groupr_temp
  $SIGZ_groupr/
@@ -103,7 +103,7 @@ stop
     NGOUT1 = 0,                         # unit for input gout tape (default=0)
     NGOUT2 = 31,                       # unit for output gout tape (default=0)
     IGG = 0,                                    # gamma group structure option
-    IWT = 11,            # weight function option (corresponding to Vitamin E)
+#    IWT = 11,            # weight function option (corresponding to Vitamin E)
     LORD = 0,                                                 # Legendre order
     NTEMP = 1,                            # number of temperatures (default=1)
     NSIGZ = 1,                            # number of sigma zeroes (default=1)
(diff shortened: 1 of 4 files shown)


tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -271,47 +295,72 @@ def set_group_structure(group_struct_arg):
...
+    if not (characteristic_parameter and parameter_name):
+        raise ValueError(
+            f'Invalid argument "{value}" for associated GROUPR parameter. ' \
+            'Consult set_modifiable_groupr_parameters() docstring for ' \
+            'guidance on accepted inputs for group structure, weight ' \
+            'function parameters.'
+        )
 
-    return ign, ngn, egn, group_name
+    return characteristic_parameter, parameter_name, ngn, egn
```
gonuke, line 360: Do we need to initialize `ngn` and `egn` as `None` for the cases where they aren't set?

tools/ALARAJOYWrapper/njoy_tools.py
```diff
@@ -271,47 +295,72 @@ def set_group_structure(group_struct_arg):
...
+        parameter_name = njoy_ref_dict[characteristic_parameter]
+
+    # Check if the provided parameter is among the list of built-in NJOY group
+    # structures or weight functions by name (values of NJOY_GROUPS or
+    # NJOY_WEIGHT_FUNCTIONS dictionaries, respectively)
+    elif value.upper() in njoy_ref_dict.values():
+        parameter_name = value.upper()
+        characteristic_parameter = list(njoy_ref_dict)[
+            list(njoy_ref_dict.values()).index(parameter_name)
         ]
```
gonuke: What if we looped through the dictionaries and tested either `value` against the key and/or the value of each entry and set accordingly?
```
for parm, name in njoy_ref_dict:
   if (value == str(parm) or value == name):
      characteristic_parameter = parm
      parameter_name = name
```
You can then do the next block only if `characteristic_parameter` is still `None`

gonuke (changes requested):
Mostly a suggestion that prevents the need for the gymnastics of dealing with the dictionary as you are now

tools/ALARAJOYWrapper/njoy_tools.py
eitan-weinstein: We have them initialized as null strings already:

```
    # Default values only needed for group structure case
    ngn = ''
    egn = ''
```

gonuke: My bad...

gonuke (approved):

PR #319, opened 2026-08-12
Print Git Commit Hash to Output

anu1217:
Currently, output from ALARA contains no record of the version used to run the program, except for the package string printed at the very top. Including the Git commit hash in the output may help with recordkeeping and debugging. Here, I'm setting up a `cmake` and header file to identify the commit hash with which ALARA is compiled.


src/alara.C
```diff
@@ -1,6 +1,6 @@
 /* $Id: alara.C,v 1.20 2004-07-29 19:24:10 wilsonp Exp $ */
 #include "alara.h"
-
+#include "git_hash.h"
```
gonuke, line 3: I don't think you need this because the value is set in `config.h` now, right?

src/CMakeLists.txt
gonuke, line 5: We can get the value of the git hash before this line, then it will be set in `config.h` without needing an additional file.

gonuke (changes requested):
I think there is a simpler way to do this by putting the git hash directly into `config.h` rather than a separate file.

PR #320, opened 2026-08-15
Modify sched_post_processor.py for Readability

anu1217:
Some minor changes are made here to `sched_post_processor.py` to improve readability and compatibility with `schedule_transforms.py` in the activationDB repository. The name of the top schedule is now stored in the schedule tree/dictionary, and the dictionary keys corresponding to the pulse and schedule delays are changed to match the format in `schedule_transforms.py`.

### DIFF
--- a/tools/sched_post_processor.py
+++ b/tools/sched_post_processor.py
@@ -34,8 +34,8 @@ def read_pulse_histories(lines):
             delay_line = lines[line_idx + 3].strip()
 
             pulse_hist_name = line.split()[1].strip("':")
-            num_pulses = num_pulse_line.split(":")[1]
-            delays = delay_line.split(":")[1]
+            num_pulses = eval(num_pulse_line.split(":")[1])
+            delays = eval(delay_line.split(":")[1])
 
             pulse_dict[pulse_hist_name] = {
                 "num_pulses_all_levels": num_pulses,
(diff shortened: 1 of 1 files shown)


tools/sched_post_processor.py
```diff
@@ -34,8 +34,8 @@ def read_pulse_histories(lines):
             delay_line = lines[line_idx + 3].strip()
 
             pulse_hist_name = line.split()[1].strip("':")
-            num_pulses = num_pulse_line.split(":")[1]
-            delays = delay_line.split(":")[1]
+            num_pulses = eval(num_pulse_line.split(":")[1])
+            delays = eval(delay_line.split(":")[1])
```
gonuke, line 38: Why `eval`?

gonuke (commented):
Curious about this choice?

tools/sched_post_processor.py
anu1217: These two lines are printed as, for example:
`num_pulses_per_level: [50]`
 `delay_seconds_per_level: [30]`
The eval was added to treat the values of these two quantities as iterables.

gonuke: So basically a way to converts string representation of an iterable to a true iterable?

gonuke (approved):

PR #325, opened 2026-08-27
Patching ALARAPlot bug for mismatching nuclides in control vs comparison runs for ratio plotting

eitan-weinstein:
This PR patches a bug in `alara_output_processing.alara_output_plotting.plot_single_response()`, wherein if the function was being called to create a ratio plot between two or more data sets, if the control (divisor) data set was missing a nuclide, a `KeyError` would be raised. This is resolved by checking against the index of the `control_piv` before calling `series /= control_piv.loc[nuc].to_numpy()` and if missing, will provide a warning and then `continue`.


gonuke (approved):
Sorry for the delay in handling all of these

PR #329, opened 2026-09-17
New PlotStats Class for ALARAPlot

eitan-weinstein:
This PR introduces a new class to `alara_output_plotting` called `PlotStats`, which is used to expand the statistical evaluation capabilities in ALARAPlot beyond just mean and standard deviation. Built in to this class are internal methods for the calculation of the following statistical metrics:

            1) Unweighted Single Series Metrics
                a) Mean (mean)
                b) Median (median)
                c) Standard Deviation (std)
                d) Variance (var)

            2) Weighted Single Series Metrics
                a) Uncertainty Weighted Mean (uncertainty-weighted-mean)

            3) Unweighted Compartive Metrics
                a) Root Mean Square Deviation (rmsd)
                b) Mean Percent Difference (mean-pct-diff)

            4) Weighted Comparative Metrics
                a) Uncertainty Weighted Root Mean Square Deviation
                   (uncertainty-weighted-rmsd)
                b) Uncertainty Weighted Mean Percent Difference
                   (uncertainty-weighted-mean-pct-diff)

The inclusion of weighted statistical metrics is undergirded by another fundamental motivation for producing these updates, which is to build in functionality to compare experimental data with simulated ADF results natively in ALARAPlot.

In addition to the actual introduction of this class to `alara_output_plotting`, this PR also includes the appropriate implementation of it to replace the old mean/standard deviation approach that the current code has.

To see this within my broader vision for experimental/computational comparative plotting and analysis, see #328.


tools/alara_output_processing/alara_output_plotting.py
```diff
@@ -7,6 +7,228 @@
 import alara_output_processing as aop
 from collections import defaultdict
 
+# ------ Comparative Series Statistics -------
+
+class PlotStats:
+    '''
+    Class containing attributes and methods used for quantitatively assessing
+        time-series plots in ALARAPlot. Can be applied for analysis simulated
```
gonuke: Something odd with indentation here...

gonuke (changes requested):
Took me a second to wrap my head around the design, in which you never explicitly instantiate a `PlotStats` object, but only implicitly when calling a class function.

PR #330, opened 2026-09-17
Functionalizing nuclide dominance range dictionary accumulation

eitan-weinstein:
This PR creates a new function in `alara_output_processing`, `deliniate_shading_regions_by_run()`, which updates the pre-existing `all_dominance_ranges` dictionary with the associated runs dominance ranges. Functionally, the code contained within `deliniate_shading_regions_by_run()` is equivalent to the code block within `plot_single_response()` that otherwise accomplished this. I was motivated to separate this to a new function both for the general principle of modularity (it accomplishes a small, distinct task) and to reapply the same functionality for the `plot_computational_with_experimental()` function introduced in #328.


gonuke (approved):
LGTM

PR #331, opened 2026-09-17
Ensuring unit consistency in ALARAPlot

eitan-weinstein:
This PR introduces a small function `get_var_unit()` to `alara_output_plotting` that ensures that all units for a single-variable ALARADFrame are the same. This will either raise a `ValueError` if they are not, or return the singular unit type. Additionally, in its implementation, each `filtered_adf` is concatenated to an accumulated ADF used for cross-checking singular unit compliance across data from _all_ runs in a single plot.

While I have not run into any issues of mismatching units, this fix seemed to me like a valuable defensive measure.

Part of the #328 changes.


gonuke (changes requested):
Do you anticipate using this elsewhere? If not, instead of keeping the filtered concatenated frame, you could just keep a set of unique units.

PR #332, opened 2026-09-17
Checking if nuclide name has already been reformatted in `reformat_isotope()`

eitan-weinstein:
This PR introduces a regular expression check within `reformat_isotope()` to see if the passed isotope name is is already in the format the function would otherwise produce. If so, then no further modifications are necessary and the isotope can be returned as-is. Without this, if a nuclide string that fit the formatting of this function were to be passed as an argument, it would have raised a `ValueError` because there would have been no `'-'` to split on.


gonuke (approved):
Trusting your regex....

